PR #251, opened 2020-10-24
Clarify that this has been added to the stdlib.

Julian:
Hi!

Minor tweak to the README opening just to mention this is already in the stdlib now yay.

Hopefully makes sense.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -2,8 +2,13 @@
  ``importlib_metadata``
 =========================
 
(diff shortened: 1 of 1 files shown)


jaraco:
This looks good. Thanks for this. I'll merge it, and then update it to capture the intention that this library also intends to support backports of changes in subsequent Python releases.

PR #252, opened 2020-10-24
ci: fix diffcov

FFY00:
Signed-off-by: Filipe Laíns <lains@archlinux.org>


jaraco:
Fixes #257.

jaraco:
Thanks for this. I was wondering why the diff-cov was failing.

PR #253, opened 2020-10-25
Fix dot handling in dist-info directory name

uranusjr:
When searching for a matching dist-info directory, try to escape dots in the name segment. This makes functions like `distribution()` able to accept PEP 503 normalized names to locate a distribution’s dist-info directory
if the distribution’s name contains dots.

Related discussions:

* pipxproject/pipx#528
* pypa/packaging.python.org#781
* https://discuss.python.org/t/clarify-naming-of-dist-info-directories/5565?u=uranusjr

### DIFF
--- a/importlib_metadata/__init__.py
+++ b/importlib_metadata/__init__.py
@@ -473,7 +473,7 @@ class FastPath:
         for child in self.children():
             n_low = child.lower()
             if (n_low in name.exact_matches
-                    or n_low.replace('.', '_').startswith(name.prefix)
+                    or n_low.startswith(name.prefix)
                     and n_low.endswith(name.suffixes)
                     # legacy case:
                     or self.is_egg(name) and n_low == 'egg-info'):
@@ -494,11 +494,18 @@ class Prepared:
         self.name = name
         if name is None:
             return
-        self.normalized = name.lower().replace('-', '_')
+        self.normalized = self.normalize(name)
         self.prefix = self.normalized + '-'
         self.exact_matches = [
             self.normalized + suffix for suffix in self.suffixes]
         self.versionless_egg_name = self.normalized + '.egg'
+
+    @staticmethod
+    def normalize(name):
+        """
+        PEP 503 normalization plus dashes as underscores.
+        """
+        return re.sub(r"[-_.]+", "-", name).lower().replace('-', '_')
 
 
 @install
--- a/importlib_metadata/__init__.py
+++ b/importlib_metadata/__init__.py
@@ -473,7 +473,7 @@ class FastPath:
         for child in self.children():
             n_low = child.lower()
             if (n_low in name.exact_matches
-                    or n_low.startswith(name.prefix)
+                    or n_low.replace('.', '_').startswith(name.prefix)
                     and n_low.endswith(name.suffixes)
                     # legacy case:
                     or self.is_egg(name) and n_low == 'egg-info'):
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -106,7 +106,7 @@ class DistInfoPkg(OnSysPath, SiteDir):
 
 class DistInfoPkgWithDot(OnSysPath, SiteDir):
     files = {
-        "pkg.dot-1.0.0.dist-info": {
+        "pkg_dot-1.0.0.dist-info": {
             "METADATA": """
                 Name: pkg.dot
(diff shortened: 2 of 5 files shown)


importlib_metadata/__init__.py
```diff
@@ -473,7 +473,7 @@ def search(self, name):
         for child in self.children():
             n_low = child.lower()
             if (n_low in name.exact_matches
-                    or n_low.startswith(name.prefix)
+                    or n_low.replace('.', '_').startswith(name.prefix)
```
jaraco, line 476: Doesn't this also need to account for other character substitutions in normalization?

tests/fixtures.py
```diff
@@ -104,6 +104,21 @@ def setUp(self):
         build_files(DistInfoPkg.files, self.site_dir)
 
 
+class DistInfoPkgWithDot(OnSysPath, SiteDir):
+    files = {
+        "pkg.dot-1.0.0.dist-info": {
```
jaraco, line 109: Since `pkg_dot` is the proposed format, I think that should be the fixture. Perhaps this fixture should also exist as a "legacy" format.

tests/test_api.py
```diff
@@ -41,6 +42,9 @@ def test_for_name_does_not_exist(self):
         with self.assertRaises(PackageNotFoundError):
             distribution('does-not-exist')
 
+    def test_for_name_containing_dot(self):
+        assert distribution('pkg-dot').metadata['Name'] == 'pkg.dot'
```
jaraco, line 46: I think the most important use-case to cover is that `distribution('pkg.dot')` should resolve when the name on the file system is normalized (`pkg_dot`). I'm not sure I want a unit test or declared support for accepting pre-normalized names.

importlib_metadata/__init__.py
jaraco: I see now this is to support legacy behaviors where egg-info and dist-info have dots.

jaraco:
In the subsequent commits, I've added tests to capture both the desired normalization behavior, but also capture support for the dot-based metadata as "legacy".

importlib_metadata/__init__.py
```diff
@@ -494,12 +494,19 @@ def __init__(self, name):
...
         self.exact_matches = [
             self.normalized + suffix for suffix in self.suffixes]
         self.versionless_egg_name = self.normalized + '.egg'
 
+    @staticmethod
+    def normalize(name):
+        """
+        PEP 503 normalization plus dashes as underscores.
+        """
+        return re.sub(r"[-_.]+", "-", name).lower().replace('-', '_')
```
uranusjr, line 508: Can this not be simplified to `re.sub(r"[-_.]+", "_", name).lower()`?

tests/fixtures.py
uranusjr: A fixture for this legacy format is definitely needed, since there are many `.dist-info` directories using it in the wild.

importlib_metadata/__init__.py
jaraco: Yes, it could. But then it becomes less clear (modifies instead of extends PEP 503). I'm divided as to which approach is more readable. Perhaps the `dist-info` docs should provide a reference implementation and this library could adopt that implementation verbatim.

tests/test_api.py
jaraco: In subsequent commits, I agree it's worthwhile to accept normalized names, though I de-emphasize that acceptance in the changelog.

tests/fixtures.py
jaraco: Confirmed and added in e7d59299d5250c286b038a1131840d76d2007e95.

jaraco:
I've targeted this PR at the maint/2.x branch so it can be rolled out for users on Python 3.5 and 2.7.

importlib_metadata/__init__.py
jaraco: Please commit whatever your preference is here. I'm happy to have something that's compatible in place and iterate more later if needed.

jaraco:
Please let me know what you think of this implementation. If you're happy enough with it, I'll merge and push it out as 2.1 and 3.1.

uranusjr:
Everything looks to me now 👍

jaraco:
FYI, this change has caused a regression in discovery of versionless distributions still using the dot (#261).

Issue #126, opened 2020-10-22
metadata object has unspecified API which leaks EmailMessage implementation


jaraco:
In GitLab by [@eschwartz](https://gitlab.com/eschwartz) on Jun 16, 2020, 15:01

Noted on IRC,

> Technically, the returned distribution metadata object is an email.message.EmailMessage instance, but this is an implementation detail, and not part of the stable API. You should only use dictionary-like methods and syntax to access the metadata contents.

According to this definition, something like `.get_all()` should not be relied upon, because it is not a dictionary method. But `.get_all()` serves a valuable role, see for example the implementation of `_read_dist_info_reqs()`, which retrieves multiple definitions of `Requires-Dist` to define the `Distribution.requires` property.

With only access to `metadata['Requires-Dist']` it's impossible to get reliable information for metadata that can have multiple values. With `metadata.items()` it's possible to get this information, but then you'd need to filter it yourself, which is clumsy and inefficient.

    <jaraco> What the project needs is: A test capturing the need for multi-valued keys, a protocol describing the interface that 'metadata' guarantees to provide, and then an update to the docs to point to the protocol as the proper interface such that EmailMessage need not be mentioned.

tl;dr it seems like a good idea to document `.get_all()` and anything else generally useful as part of the public `Distribution.metadata`, API, without mentioning `EmailMessage` at all.

jaraco:
In GitLab by [@FFY00](https://gitlab.com/FFY00) on Jun 16, 2020, 15:39

I believe `EmailMessage` should still be mentioned as otherwise people will assume the other methods are just missing from the documentation.

jaraco:
In GitLab by [@jaraco](https://gitlab.com/jaraco) on Sep 22, 2020, 20:13

In 5e7d0d8, I've attempted to add an interface definition on the return type for metadata. Unfortunately, it causes this error from mypy:

```
importlib_metadata/__init__.py:287: error: Cannot call function of unknown type
```

It seems the `email_message_from_string` function is an unknown type. The [definition for that function](https://gitlab.com/python-devs/importlib_metadata/-/blob/46efbf9df587e0a799fa6166dbc056486e02d106/importlib_metadata/_compat.py#L107-118) is selective on Python version, so I'm not even sure mypy can discern the return type for the different forms that function might take. I tried placing a `cast(PackageMetadata, email_message_from_string(text))`, but that resulted in the same error.

Does anyone with more familiarity with mypy know how to declare the types in this situation?

[...]

Issue #257, opened 2020-11-22
Diffcov failing since migration to Github Actions


jaraco:
Diffcov tests are failing thus:

```
diffcov run-test: commands[2] | python -m coverage combine --rcfile=/home/runner/work/importlib_metadata/importlib_metadata/coverage.ini
diffcov run-test: commands[3] | python -m coverage xml --rcfile=/home/runner/work/importlib_metadata/importlib_metadata/coverage.ini
diffcov run-test: commands[4] | diff-cover coverage.xml --html-report diffcov.html
Traceback (most recent call last):
  File "/home/runner/work/importlib_metadata/importlib_metadata/.tox/diffcov/bin/diff-cover", line 8, in <module>
    sys.exit(main())
  File "/home/runner/work/importlib_metadata/importlib_metadata/.tox/diffcov/lib/python3.9/site-packages/diff_cover/diff_cover_tool.py", line 216, in main
    percent_covered = generate_coverage_report(
  File "/home/runner/work/importlib_metadata/importlib_metadata/.tox/diffcov/lib/python3.9/site-packages/diff_cover/diff_cover_tool.py", line 184, in generate_coverage_report
    reporter.generate_report(output_file)
  File "/home/runner/work/importlib_metadata/importlib_metadata/.tox/diffcov/lib/python3.9/site-packages/diff_cover/report_generator.py", line 251, in generate_report
    report = template.render(self._context())
  File "/home/runner/work/importlib_metadata/importlib_metadata/.tox/diffcov/lib/python3.9/site-packages/diff_cover/report_generator.py", line 300, in _context
    context = super().report_dict()
  File "/home/runner/work/importlib_metadata/importlib_metadata/.tox/diffcov/lib/python3.9/site-packages/diff_cover/report_generator.py", line 183, in report_dict
    src_stats = {src: self._src_path_stats(src) for src in self.src_paths()}
  File "/home/runner/work/importlib_metadata/importlib_metadata/.tox/diffcov/lib/python3.9/site-packages/diff_cover/report_generator.py", line 85, in src_paths
    for src, summary in self._diff_violations().items()
  File "/home/runner/work/importlib_metadata/importlib_metadata/.tox/diffcov/lib/python3.9/site-packages/diff_cover/report_generator.py", line 178, in _diff_violations
    for src_path in self._diff.src_paths_changed()
[...]

PR #263, opened 2020-12-01
importlib_metadata: define protocol for Distribution.metadata

FFY00:

### DIFF
--- a/.github/workflows/automerge.yml
+++ b/.github/workflows/automerge.yml
@@ -1,0 +1,27 @@
+name: automerge
+on:
+  pull_request:
+    types:
+      - labeled
+      - unlabeled
+      - synchronize
+      - opened
+      - edited
+      - ready_for_review
+      - reopened
+      - unlocked
+  pull_request_review:
+    types:
+      - submitted
+  check_suite:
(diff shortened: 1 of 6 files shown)


FFY00:
Resolves #126.

FFY00:
We run into the case here where we have different coverage in different versions, because of the backport import that will only run on older Python. The proper solution would be to run tests for both versions and merge the coverage, but since it is just an import, I added a `# pragma: no cover` there.

FFY00:
Btw, relevant typeshed definitions: https://github.com/python/typeshed/blob/master/stdlib/3/email/message.pyi

jaraco:
This looks great. I'll add a changelog entry and get it out. Thanks!

jaraco:
Well drat. I renamed the default branch from `master` to `main` and now diffcov is failing again as it was in #257, despite the fix still being applied. It seems the name 'master' is cached or hard-coded in diff-cover.

jaraco:
I'm going to proceed with manually merging and ignore the diff-cover failure for now. If the issue persists in new PRs, I'll probably just get rid of that footgun.

Issue #97, opened 2020-10-22
pypy2: EntryPoint object repr results in infinite recursion


jaraco:
In GitLab by [@asottile](https://gitlab.com/asottile) on Nov 29, 2019, 18:08

```
>>> from importlib_metadata import EntryPoint
>>> EntryPoint('foo', 'bar', None)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>

...

  File "<string>", line 22, in __repr__
  File "<string>", line 22, in __repr__
  File "<string>", line 22, in __repr__
  File "<string>", line 22, in __repr__
  File "<string>", line 22, in __repr__
  File "/home/asottile/workspace/flake8/.tox/pypy/site-packages/importlib_metadata/__init__.py", line 125, in __iter__
    return iter((self.name, self))
RuntimeError: maximum recursion depth exceeded
```

jaraco:
In GitLab by [@asottile](https://gitlab.com/asottile) on Nov 29, 2019, 18:17

mentioned in commit asottile/flake8@6b764d0ec7fbbc3ecfe00d9a3705a4fc9934e7b4

jaraco:
In GitLab by [@asottile](https://gitlab.com/asottile) on Nov 29, 2019, 18:18

mentioned in merge request pycqa/flake8!389

jaraco:
In GitLab by [@asottile](https://gitlab.com/asottile) on Nov 30, 2019, 02:33

I'd be interested in working on this, the root cause is the same as #96 -- generally when extending `NamedTuple` overriding `__iter__` is a bad idea (since things expecting tuples will expect it to work in a certain way).  In this case, one of the other tuple member functions makes assumptions about `__iter__` (which are violated by the specification in `importlib.metadata`s `EntryPoint`

I'm wondering if at this point it's too late to back out `__iter__` from `EntryPoint` (it's provisional in python3.8 right?)

jaraco:
In GitLab by [@jaraco](https://gitlab.com/jaraco) on Nov 30, 2019, 10:24

mentioned in commit 10234d54b6f72e7c7a404e6324fb12c69dbaff86

jaraco:
In GitLab by [@jaraco](https://gitlab.com/jaraco) on Nov 30, 2019, 10:25

In the aforementioned commit, I added a test. The test only fails on PyPy 2, but passes on PyPy 3.

jaraco:
[...]

PR #278, opened 2021-01-23
Feature/entry points by group and name

jaraco:
In pypa/twine#728, I learned that the `dict` hack for resolving a group of entry points by name is unintuitive, violates static type checks, and still doesn't provide an elegant way to resolve a set of entry points by group and name.

This finding inspired me to create this patch to provide an even nicer experience. See the changes to the docs and tests for examples of the improved usage.

The change is backward compatible for usages under test, but deprecates the `__iter__` on EntryPoint and `dict`-based construction.

### DIFF
--- a/importlib_metadata/__init__.py
+++ b/importlib_metadata/__init__.py
@@ -130,10 +130,6 @@ class EntryPoint(
         config.read_string(text)
         return cls._from_config(config)
 
-    @classmethod
-    def _from_text_for(cls, text, dist):
-        return (ep._for(dist) for ep in cls._from_text(text))
-
     def _for(self, dist):
         self.dist = dist
         return self
@@ -155,34 +151,41 @@ class EntryPoint(
             (self.name, self.value, self.group),
         )
 
+    def matches(self, **params):
+        attrs = (getattr(self, param) for param in params)
+        return all(map(operator.eq, params.values(), attrs))
+
 
 class EntryPoints(tuple):
     """
-    An immutable collection of EntryPoint objects, retrievable by name.
+    An immutable collection of selectable EntryPoint objects.
     """
 
     __slots__ = ()
 
-    def __getitem__(self, name) -> EntryPoint:
+    def __getitem__(self, name) -> Union[EntryPoint, 'EntryPoints']:
         try:
-            return next(ep for ep in self if ep.name == name)
-        except Exception:
+            match = next(iter(self.select(name=name)))
+            return match
+        except StopIteration:
+            if name in self.groups:
+                return self._group_getitem(name)
             raise KeyError(name)
+
+    def _group_getitem(self, name):
+        """
+        For backward compatability, supply .__getitem__ for groups.
+        """
+        msg = "GroupedEntryPoints.__getitem__ is deprecated for groups. Use select."
+        warnings.warn(msg, DeprecationWarning)
+        return self.select(group=name)
+
+    def select(self, **params):
+        return EntryPoints(ep for ep in self if ep.matches(**params))
 
     @property
     def names(self):
         return set(ep.name for ep in self)
-
-
-class GroupedEntryPoints(tuple):
-    """
-    An immutable collection of EntryPoint objects, retrievable by group.
-    """
-
-    __slots__ = ()
-
-    def __getitem__(self, group) -> EntryPoints:
-        return EntryPoints(ep for ep in self if ep.group == group)
 
     @property
     def groups(self):
@@ -193,9 +196,13 @@ class GroupedEntryPoints(tuple):
         For backward compatibility, supply .get
         """
         is_flake8 = any('flake8' in str(frame) for frame in inspect.stack())
-        msg = "GroupedEntryPoints.get is deprecated. Just use __getitem__."
+        msg = "GroupedEntryPoints.get is deprecated. Use select."
         is_flake8 or warnings.warn(msg, DeprecationWarning)
-        return self[group] or default
+        return self.select(group=group) or default
+
+    @classmethod
+    def _from_text_for(cls, text, dist):
+        return cls(ep._for(dist) for ep in EntryPoint._from_text(text))
 
 
 class PackagePath(pathlib.PurePosixPath):
@@ -353,8 +360,7 @@ class Distribution:
 
     @property
     def entry_points(self):
-        eps = EntryPoint._from_text_for(self.read_text('entry_points.txt'), self)
-        return GroupedEntryPoints(eps)
+        return EntryPoints._from_text_for(self.read_text('entry_points.txt'), self)
 
     @property
     def files(self):
@@ -687,13 +693,13 @@ def version(distribution_name):
     return distribution(distribution_name).version
 
 
-def entry_points():
+def entry_points(**params):
     """Return EntryPoint objects for all installed packages.
 
     :return: EntryPoint objects for all installed packages.
     """
     eps = itertools.chain.from_iterable(dist.entry_points for dist in distributions())
-    return GroupedEntryPoints(eps)
+    return EntryPoints(eps).select(**params)
 
 
 def files(distribution_name):
--- a/importlib_metadata/__init__.py
+++ b/importlib_metadata/__init__.py
@@ -180,6 +180,11 @@ class EntryPoints(tuple):
 
     @property
     def groups(self):
+        """
+        For coverage while SelectableGroups is present.
+        >>> EntryPoints().groups
+        set()
+        """
         return set(ep.group for ep in self)
 
     @classmethod
@@ -202,11 +207,16 @@ class SelectableGroups(dict):
 
     @property
     def groups(self):
-        return self.keys()
+        return set(self.keys())
 
     @property
     def names(self):
-        return (ep.name for ep in self._all)
+        """
+        for coverage:
+        >>> SelectableGroups().names
+        set()
+        """
+        return set(ep.name for ep in self._all)
 
     @property
     def _all(self):
@@ -218,7 +228,7 @@ class SelectableGroups(dict):
         return EntryPoints(self._all).select(**params)
 
 
-class LegacyGroupedEntryPoints(EntryPoints):
+class LegacyGroupedEntryPoints(EntryPoints):  # pragma: nocover
     """
     Compatibility wrapper around EntryPoints to provide
     much of the 'dict' interface previously returned by
--- a/importlib_metadata/__init__.py
+++ b/importlib_metadata/__init__.py
@@ -183,6 +183,14 @@ class GroupedEntryPoints(tuple):
     def groups(self):
         return set(ep.group for ep in self)
 
+    def get(self, group, default=None):
+        """
+        For backward compatibility, supply .get
+        """
+        msg = "GroupedEntryPoints.get is deprecated. Just use __getitem__."
+        warnings.warn(msg, DeprecationWarning)
+        return self[group] or default
+
 
 class PackagePath(pathlib.PurePosixPath):
     """A reference to a path in a package"""
--- a/importlib_metadata/__init__.py
+++ b/importlib_metadata/__init__.py
@@ -165,22 +165,11 @@ class EntryPoints(tuple):
 
     __slots__ = ()
 
-    def __getitem__(self, name) -> Union[EntryPoint, 'EntryPoints']:
+    def __getitem__(self, name):  # -> EntryPoint:
         try:
-            match = next(iter(self.select(name=name)))
-            return match
+            return next(iter(self.select(name=name)))
         except StopIteration:
-            if name in self.groups:
-                return self._group_getitem(name)
             raise KeyError(name)
-
-    def _group_getitem(self, name):
-        """
-        For backward compatability, supply .__getitem__ for groups.
-        """
-        msg = "GroupedEntryPoints.__getitem__ is deprecated for groups. Use select."
-        warnings.warn(msg, DeprecationWarning)
-        return self.select(group=name)
 
     def select(self, **params):
         return EntryPoints(ep for ep in self if ep.matches(**params))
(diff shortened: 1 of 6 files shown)


jaraco:
> The change is backward compatible for usages under test,

[Hyrum's law](https://www.hyrumslaw.com/) strikes again. Seems Flake8 is relying on the output of `entry_points()` being a proper dict and having a `.get` method.

jaraco:
> Seems Flake8 is relying on the output of `entry_points()` being a proper dict and having a `.get` method.

So this finding begs the question, what other use-cases are there out there that expect `entry_points` to return a mutable mapping? Are there use-cases that rely on `__iter__` to return the group names? Are there use-cases that mutate the dict after? Are those use-cases dominant enough to demand backward-compatibility and deprecation warnings (like I've done for `.get` in 342a94b?

jaraco:
Here's my answer to those questions - I recommend to merge this change as written, cut a release of `importlib_metadata`, and see if users report any issues. If not, then it's probably safe to merge into CPython 3.10. Otherwise, I'll add compatibility shims to support the reported issues as well.

jaraco:
@sigmavirus24: Would you care to review this change, given that it was inspired by a `twine` issue and affects `flake8`?

jaraco:
I plan to merge this as early as Friday unless I hear from Barry or Ian that they'd like me to hold off and await their review.

sigmavirus24:
@asottile you rewrote flake8 to use importlib-metadata. Thoughts?

importlib_metadata/__init__.py
```diff
@@ -153,6 +155,43 @@ def __reduce__(self):
...
+    @property
+    def groups(self):
+        return set(ep.group for ep in self)
+
+    def get(self, group, default=None):
+        """
+        For backward compatibility, supply .get
+        """
+        msg = "GroupedEntryPoints.get is deprecated. Just use __getitem__."
+        warnings.warn(msg, DeprecationWarning)
```
asottile, line 191: seems odd to immediately deprecate this when there's no way to write the code which is compatible with the existing versions (`__getitem__` doesn't work for old importlib.metadata or importlib_metadata)

this is going to lead to a bunch of people reporting DeprecationWarning and flake8 being able to do nothing about it

importlib_metadata/__init__.py
```diff
@@ -153,6 +155,43 @@ def __reduce__(self):
...
+        return set(ep.name for ep in self)
+
+
+class GroupedEntryPoints(tuple):
+    """
+    A collection of EntryPoint objects, retrievable by group.
+    """
+
+    def __getitem__(self, group) -> EntryPoints:
+        return EntryPoints(ep for ep in self if ep.group == group)
```
asottile, line 180: A bit unfortunate that this is `O(N)` on every key access -- this will likely surprise consumers who may access this in a loop

importlib_metadata/__init__.py
```diff
@@ -153,6 +155,43 @@ def __reduce__(self):
         )
 
 
+class EntryPoints(tuple):
```
asottile, line 158: the pickle and json shape of both of these objects changes in a breaking way

additionally, I usually find it's best to [set `__slots__ = ()` when extending `tuple` to preserve immutability](https://youtu.be/BSNd_kxHXL8)

importlib_metadata/__init__.py
jaraco: There are a few things flake8 could do:

- Instead of using `.get`, construct a dictionary explicitly from items (`eps = {ep.name: ep for ep in entry_points()}`). That method will work on old and new importlib_metadata.
- Suppress the deprecation warning.
- (my preference) just bump the requirement to the new release of `importlib_metadata` and use `__getitem__`.

importlib_metadata/__init__.py
jaraco: Agreed. Perhaps there's a more efficient way to handle it. If there's a use-case that demands it, I'd be happy to investigate an optimization. I do expect this approach to be less complex than sorting and grouping by group as before.

importlib_metadata/__init__.py
jaraco: Pickle and JSON shape of entry point containers isn't a tested or documented interface. Do you think it should be? My instinct is it's YAGNI or at least an unsupported expectation.

Good suggestion on `__slots__`. That seems like a good idea. Usually, I prefer a minimal implementation, but in this case, it seems reasonable.

importlib_metadata/__init__.py
asottile: in case it wasn't clear -- I'm asking you please do not immediately deprecate something that's currently in use.  it will cause me a lot of pain dealing with users that can't read.

importlib_metadata/__init__.py
asottile: it isn't tested or documented, but since it is part of the standard library you'll want to adhere to best practices for cpython which essentially boils down to "if it could be json'd or pickled, people are already doing that and changing that is a breaking change" -- or at least that has been my experience working with contributions to cpython

jaraco: In relation to CPython, this change will go into 3.10, and pickle compatibility and shape is not guaranteed across minor releases. Moreover, I'm struggling how to think to improve the situation without breaking that expectation. Any introduction of a new class will automatically be incompatible on prior versions (unable to deserialize there). I appreciate the feedback, but I think this is a necessary breakage.

importlib_metadata/__init__.py
jaraco: I've never had such a request before, so I don't have any prior experience from which to draw. Part of the reason I engaged with the flake8 project was because of its prominence and my desire to address proactively the impact of this change on that project. How long should a deprecation warning be suppressed? Just until a subsequent release? A fixed period of time? Would a PendingDeprecationWarning be better for the interim or is any warning as bad as any other?

asottile: other things in cpython seem to go:

- introduce new thing and don't warn in old thing until N releases pass (usually until the latest one with the old way is unsupported)
- PendingDeprecationWarning (>= 1 release)
- DeprecationWarning (>= 1 release)
- removed

your recommendation of bumping the version of `importlib_metadata` is very undesirable as it would require *re-introducing* a dependency in python3.8+ -- just to hack around this

jaraco: In eb170ba, I've put a 60 day delay on the deprecation warnings. Will that do the trick?

asottile:
and actually, taking a step back you argued exactly the opposite point when I pointed this out originally (that it *can't be removed*): https://github.com/python/importlib_metadata/issues/97#issuecomment-714767479

importlib_metadata/__init__.py
asottile: no it will not, flake8 releases about once a year

importlib_metadata/__init__.py
```diff
@@ -647,10 +699,7 @@ def entry_points():
     :return: EntryPoint objects for all installed packages.
     """
     eps = itertools.chain.from_iterable(dist.entry_points for dist in distributions())
-    by_group = operator.attrgetter('group')
-    ordered = sorted(eps, key=by_group)
-    grouped = itertools.groupby(ordered, by_group)
-    return {group: tuple(eps) for group, eps in grouped}
```
asottile, line 653: if you back out this portion of the patch I have no opposition -- keeping this a dictionary makes a lot of common operations easier

importlib_metadata/__init__.py
```diff
@@ -308,7 +359,8 @@ def version(self):
 
     @property
     def entry_points(self):
-        return list(EntryPoint._from_text_for(self.read_text('entry_points.txt'), self))
+        eps = EntryPoint._from_text_for(self.read_text('entry_points.txt'), self)
+        return GroupedEntryPoints(eps)
```
asottile, line 363: I believe this changes the type from `list` to a mapping-like -- this doesn't seem correct?

importlib_metadata/__init__.py
jaraco: I'm trying to serve two use-cases here, the very common access by group, but the also not uncommon access to all entrypoints irrespective of group. I see the list of all entry points as the lowel-level interface and the mapping as a more useful but higher-level interface. By providing only the higher-level interface, it becomes clumsy and expensive to derive the list of all entry points from the mapping of entry points by group. It also becomes impossible to restore the original order of entrypoints. One use-case I'd like to address (I forget where it was requested) is to allow a tool to simply enumerate all entry points.

If we're going to go to the trouble of tweaking the interface, I'd like to make it more general and useful and regular.

By wrapping the result in a GroupedEntryPoints, one gets the generality of the list of entry points but the convenience of accessing them by group (like the dict construction allowed). Moreover, users are given an explicit interface to support these use-cases rather inspect the code to find an approach that happens to work (e.g. `all_entry_points = list(itertools.chain.from_iterable(entry_points().values()))`).

Moreover, `GroupedEntryPoints` has the additional advantage of supplying the empty set when no packages present the group, making flake8's usage more straightforward as well:

```diff
- eps = importlib_metadata.entry_points().get(self.namespace, ())
+ eps = importlib_metadata.entry_points()[self.namespace]
```

importlib_metadata/__init__.py
jaraco: It changes from a list to a GroupedEntryPoints, which is a tuple subclass with `__getitem__` for easy access by group (and no access by index). It's a little magical, but well within the norms of OO programming.

importlib_metadata/__init__.py
asottile: to be clear, I don't care if the resulting code "looks cleaner".  if I have to deal with users complaining about a DeprecationWarning because you decided an api can be more clever I'm not happy.

these cutesy types that half imitate a dict but aren't a dict and a list but are neither *are going to confuse users* and by switching to them you cause me and my users pain

importlib_metadata/__init__.py
jaraco: I want this deprecation warning to go into Python 3.10. I agree there are some cases where deprecations happen much slower, but this case has a workaround that works across versions _and_ a backport (this library), so the motivations for a long deprecation period aren't present.

jaraco:
> and actually, taking a step back you argued exactly the opposite point when I pointed this out originally (that it _can't be removed_): [#97 (comment)](https://github.com/python/importlib_metadata/issues/97#issuecomment-714767479)

I'm not following this concern exactly (I'm unsure what point was argued). I re-read that thread from that comment and it seems that I was not adamant about the approach, but ultimately settled on the current compromise that I'd now like to amend. At the time, I didn't have another proposal that met the goals, but I believe this approach does. Your concerns were relevant then and are now. If you have a concern that's not addressed in another thread, could you re-state it and what you recommend?

importlib_metadata/__init__.py
jaraco: In c9adbca, I've added an escape hatch for flake8.

jaraco: (amended with 71fd4a7, because the prior commit was broken)

jaraco:
I've been considering an alternative experience that uses parameters to `entry_points` to solicit the `group` and `name`. In particular, something like:

```
def entry_points(group=All, name=All):
    return (ep for ep in dist.entry_points() for dist in distributions() if ep.name == name and ep.group == group)
```

In this form, the user will always get an iterator of EntryPoint objects and they can filter on group or name but otherwise just get an iterator of EntryPoints. The result would need to be wrapped in some object to provide backward compatibility, but that wrapper would eventually be removed.

A materialized list could be used in place of an iterator, although that approach would give the caller less control over the execution.

Would that approach be preferable?

jaraco:
> I've been considering an alternative experience that uses parameters to `entry_points` to solicit the `group` and `name`.

This approach doesn't work, because when one passes `name`, they expect to get a single EntryPoint back, but when they don't pass `name`, they'd like to get an iterable of EntryPoints, possibly keyed by name. If the function always returns an iterable or sequence, the user is forced to unpack it. Maybe then `name=` should be omitted.

jaraco:
Omitting `name=` from the query doesn't help... because there are still two primary ways that users wish to access the results - either as a single entry point resolved by name or as a sequence of entry points. Since I'd like to avoid users requiring to import multiple names for this small set of common cases, I'm driven back to creating a `EntryPoints` object that allows resolving an EntryPoint by name from a set of EPs. At which point, I ask myself, can the groups of EntryPoints follow the same API for simplicity and uniformity, which leads me back to the original proposal above.

The whole point of this library is to provide imminently intuitive interfaces for the most common use-cases. Providing dumb objects that require copy/paste from docs or other examples or requiring importing manipulator functions/adapters for common use-cases would be a fail.

I'm struggling to come up with a meaningfully better approach that doesn't compromise the use cases. Unless someone else can, I'm likely to move forward with this approach or something very similar.

asottile:
a quick search on github shows that hundreds of others are using the returned dictionary and return list as dictionaries and lists -- please don't break them

I'm unsubscribing this thread since you're not listening to me, if you want my opinion again please @-mention me, but my current opinion about this is a big thumbs down and I think you're making things worse.  when this gets to beta review I'll be sure to bring this pain up with the release manager as well -- I really don't want the same pain that happened with 3.9 to happen again because of importlib metadata

jaraco:
> I'm unsubscribing this thread since you're not listening to me.

With all due respect, I am listening, and I care very much about your opinion, and I want to find a solution and preferably one that doesn't take several years to implement. Thusfar, your recommendations have primarily been "don't change anything," which is why I've focused on explaining the motivations for the change. I've since added more context in separate issues to help distill the motivations and elucidate the challenges. I'm sorry you've disengaged, but I'll try to push forward without you and honor the concerns you've raised.

> a quick search on github shows that hundreds of others are using the returned dictionary and return list as dictionaries and lists -- please don't break them

I believe that's what I've done. The current approach retains the `__getitem__` and `get` support and acknowledges there may be other uses that could break. I'd love to be able to discover those early and see if a design could be implemented to provide compatibility. I wish you'd saved the query or provided some characteristic references. I tried doing the search myself and found it very difficult to find meaningful results.

My plan currently is to come up with a design that's largely compatible, has easy adjustments for rare incompatibilities, and declares this change with a backward-incompatible release.

importlib_metadata/__init__.py
jaraco: The latest code no longer tries to half imitate a dict, but simply uses `__getitem__` for entry point access by name.

jaraco:
I've been thinking about the compatibility concerns and I have a new idea - it should be possible to cut a fully compatible release of these changes, exposing the new `.select` behavior, and then defer the backward-incompatible behavior for a future release.

jaraco:
In this latest version, I've come up with a fully compatible release with forward compatibility for the new, simpler, and more flexible interface. I'd like to get it out to ensure it's stable. Thanks for the comments - they really helped me figure out a better transition here.

PR #281, opened 2021-01-27
Restrict to unique distributions in entry points

### LINKED ISSUE #280: Metadata returned even for superseded packages
In jaraco/keyring#486, the reporter encountered an issue when keyring 22 changed the name of an entry point. After the change, with the latest version installed with `--user` and an older version installed in system site-packages, there are two copies of keyring installed, the latest version taking precedence. `importlib_metadata`, however, loads metadata for _both_ packages, including references to the unreachable and no longer valid entry points from the older package.

Probably the `entry_points` loading mechanism should only return entry points for the first package by any given name.

jaraco:
Closes #280.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -3,6 +3,7 @@ omit =
 	*/.tox/*
 	tests/*
 	prepare/*
+	*/_itertools.py
 
 [report]
 show_missing = True
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,9 @@
+v3.5.0
+======
+
+* #280: ``entry_points`` now only returns entry points for
+  unique distributions (by name).
+
 v3.4.0
 ======
 
--- a/importlib_metadata/__init__.py
+++ b/importlib_metadata/__init__.py
@@ -19,6 +19,8 @@ from ._compat import (
     Protocol,
 )
 
+from ._itertools import unique_everseen
+
 from configparser import ConfigParser
 from contextlib import suppress
 from importlib import import_module
@@ -646,7 +648,10 @@ def entry_points():
 
     :return: EntryPoint objects for all installed packages.
     """
-    eps = itertools.chain.from_iterable(dist.entry_points for dist in distributions())
+    unique = functools.partial(unique_everseen, key=operator.attrgetter('name'))
+    eps = itertools.chain.from_iterable(
+        dist.entry_points for dist in unique(distributions())
+    )
     by_group = operator.attrgetter('group')
     ordered = sorted(eps, key=by_group)
     grouped = itertools.groupby(ordered, by_group)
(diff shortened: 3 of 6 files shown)


jaraco:
In this latest commit, I add 2b64fa2 against the main branch to test performance of entry_points before the patch and now after. On my 2019 Macbook Pro, I see 9.5ms as the baseline and 28ms after the patch, a 200% increase.

anntzer:
I'd guess the way to avoid the large performance penalty is along the lines of (simplified here)
```patch
diff --git i/importlib_metadata/__init__.py w/importlib_metadata/__init__.py
index 4c420a5..2a90b6a 100644
--- i/importlib_metadata/__init__.py
+++ w/importlib_metadata/__init__.py
@@ -590,6 +590,7 @@ class PathDistribution(Distribution):
                      .joinpath(), __div__, .parent, and .read_text().
         """
         self._path = path
+        self._normalized_name = path.stem.partition("-")[0]
 
     def read_text(self, filename):
         with suppress(
@@ -648,7 +649,7 @@ def entry_points():
 
     :return: EntryPoint objects for all installed packages.
     """
-    unique = functools.partial(unique_everseen, key=operator.attrgetter('name'))
+    unique = functools.partial(unique_everseen, key=operator.attrgetter('_normalized_name'))
     eps = itertools.chain.from_iterable(
         dist.entry_points for dist in unique(distributions())
     )
```
(on top of your PR)
i.e. take advantage that in practice, the vast majority of Distributions will be PathDistributions, for which the normalized name can be directly inferred from the filename (I didn't handle eggs here, but something similar applies, of course modulo the different normalization).  To handle non-PathDistributions, the base Distribution class should also gain a `_normalized_name` property that computes it from the non-normalized name.
Of course, this is assuming that you distributions with different names but with the same normalized name can be considered identical; if that's not the case you'd need something slightly more elaborate but along the same lines (first key by normalized name, then for each group of multiple distributions with the same normalized name, if that group has size>1 (which should be rare in practice), actually resolve the non-normalized names and do the deduplication).
The patch above (which doesn't implement all these subtleties) removes nearly all the overhead of this PR.

domdfcoding:
> this is assuming we indeed don't need to support interpolation or anything fancy

Could it fall back to ConfigParser in those cases?

anntzer:
> Could it fall back to ConfigParser in those cases?

Yes, but we'd need to actually detect if interpolation is being used (for which I guess `"%(" in blah` is a close enough approximation with no false negatives...).

jaraco:
I'm all but certain that interpolation isn't expected. In fact, [that's how pkg_resources parses it](https://github.com/pypa/setuptools/blob/a4bb2ca77ab379afe3e254781fdedad389d824e2/pkg_resources/__init__.py#L3180-L3203). So it's definitely safe not to rely on ConfigParser.

For this PR, I'm going to accept it as written, and we can work on optimization techniques subsequently.

Issue #292, opened 2021-03-14
Optimize regex compilation in Distribution._read_sections


tirkarthi:
In below lines the regex pattern is compiled for each line in a loop which is not needed since we can just use `re.compile` once and use the pattern object. This improves the speed as per below benchmarks.

https://github.com/python/importlib_metadata/blob/5c9198cd8212d56db8ee9c49ee39a0980d11f5ff/importlib_metadata/__init__.py#L528-L535

Benchmark of comparing requires.txt for dask

```
$  wc requires.txt
 38  31 427 requires.txt
$ python -m pyperf timeit -s 'from importlib_metadata import Distribution; requires = open("requires.txt").read()' 'Distribution._deps_from_requires_text(requires)' -o no_pattern.json
$ python -m pyperf timeit -s 'from importlib_metadata import Distribution; requires = open("requires.txt").read()' 'Distribution._deps_from_requires_text(requires)' -o pattern.json
$ python -m pyperf compare_to no_pattern.json pattern.json --table
+-----------+------------+-----------------------+
| Benchmark | no_pattern | pattern               |
+===========+============+=======================+
| timeit    | 53.4 us    | 30.4 us: 1.76x faster |
+-----------+------------+-----------------------+
```

jaraco:
> the regex pattern is compiled for each line

That's surprising to me. I thought regex cached these patterns.

> Benchmark of comparing requires.txt for dask

Oh, today I learned of pyperf. Thanks for the reference. I'd really like to get this functionality implemented as a 'compare against main' feature. Thanks for sharing.

jaraco:
I note that as of 3.9, this module has two implementations of a config parser, one in the Sectioned class, and the other in the _read_sections method. Probably those should be consolidated.

tirkarthi:
I looked into implementation n I can see the caching at https://github.com/python/cpython/blob/af50c84643ce21cfbdfdabbdfae6bd5e1368c542/Lib/re.py#L285 . It has max entries of 512. I am not sure if this is due to the way the benchmark is done in pyperf or timeit. I don't guess the cache limit is reached so still little confused by benchmark result as well.

jaraco:
After merging #294, I set out to create a test that would capture the performance concern implied above.

[...]

PR #294, opened 2021-03-28
Employ pyperf and separate environments to compare performance

jaraco:
Ref #292.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -35,7 +35,7 @@ jobs:
       - name: Run benchmarks
         run: tox
         env:
-          TOXENV: perf
+          TOXENV: perf{,-ref}
 
   diffcov:
     runs-on: ubuntu-latest
--- a/tox.ini
+++ b/tox.ini
@@ -33,19 +33,44 @@ commands =
 	diff-cover coverage.xml --compare-branch=origin/main --html-report diffcov.html
 	diff-cover coverage.xml --compare-branch=origin/main --fail-under=100
 
-[testenv:perf]
-use_develop = False
+[perf]
 deps =
 	ipython
+	pyperf
+	path
 commands =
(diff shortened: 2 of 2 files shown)


jaraco:
Annoyingly, even with no code changes, pyperf is reporting a statistically-significant performance difference:

```
perf-ref run-test: commands[5] | python -m pyperf compare_to --verbose main.json local.json --table
+-----------------------+---------+-----------------------+
| Benchmark             | main    | local                 |
+=======================+=========+=======================+
| entry_points()        | 7.13 ms | 23.6 ms: 3.30x slower |
+-----------------------+---------+-----------------------+
| uncached distribution | 806 us  | 1.33 ms: 1.65x slower |
+-----------------------+---------+-----------------------+
| Geometric mean        | (ref)   | 1.53x slower          |
+-----------------------+---------+-----------------------+

Benchmark hidden because not significant (2): discovery, cached distribution
```

I noticed the same thing on my local machine:

```
perf-ref run-test: commands[5] | python -m pyperf compare_to --verbose main.json local.json --table
+-----------------------+---------+-----------------------+
| Benchmark             | main    | local                 |
+=======================+=========+=======================+
| entry_points()        | 6.11 ms | 19.1 ms: 3.13x slower |
+-----------------------+---------+-----------------------+
| cached distribution   | 101 us  | 99.6 us: 1.02x faster |
+-----------------------+---------+-----------------------+
| uncached distribution | 891 us  | 1.29 ms: 1.45x slower |
+-----------------------+---------+-----------------------+
| Geometric mean        | (ref)   | 1.45x slower          |
+-----------------------+---------+-----------------------+

Benchmark hidden because not significant (1): discovery
```

And since the two environments are essentially identical, I'm really surprised.

Perhaps there's some way to tune the system.

As it is, this technique is barely better than the current system and is in some ways worse, because it's presenting untrustworthy summaries.

PR #304, opened 2021-04-18
Repair and clean metadata

### LINKED ISSUE #275: Description not available through PackageMetadata protocol
While investigating the details of Python packaging metadata for [a Setuptools issue](https://github.com/pypa/setuptools/issues/1390#issuecomment-762410744), I learned that [PEP 566](https://www.python.org/dev/peps/pep-0566/) allows the `Description` to be indicated as a field or as the payload of the metadata and in fact the latter approach is what is seen in dist-info metadata as produced by today's tools.

That means that the [PackageMetadata Protocol](https://github.com/python/importlib_metadata/blob/bde4e157b4ac1924a953897c92ed0b2638cdb229/importlib_metadata/__init__.py#L179-L192) does not expose that information. For example:

```
draft $ pip-run -q setuptools
>>> import importlib.metadata
>>> dist = importlib.metadata.distribution('setuptools')
>>> dist.metadata['Description'] or 'missing'
'missing'
```

But the content is there:

```
>>> dist.metadata.get_payload()[:10]
'.. image::'
```

Probably importlib.metadata could copy that payload into the Description field, but it probably needs to first [encode the content according to this spec](https://www.python.org/dev/peps/pep-0345/#id17) (yuck).

### LINKED ISSUE #276: Support loading of multi-line fields
As discovered in [this comment](https://github.com/pypa/setuptools/issues/1390#issuecomment-762410744), at least some of the fields support multi-line values with different mechanisms for encoding the continuation from previous lines. It would be nice if the `PackageMetadata` object returned by `Distribution.metadata` had support for decoding these lines to their essential values (removing indentation and continuation characters (`|` in the case of Description)).

### LINKED ISSUE #277: Support JSON encoding of metadata
PEP 566 provides a [specification for decoding the metadata to JSON](https://www.python.org/dev/peps/pep-0566/#json-compatible-metadata). Importlib.metadata should probably implement that transform. Annoyingly, as written, that spec would require the JSON-transformed form to include RFC822 artifacts (padding, continuation characters) in multi-line values.

jaraco:
Superseding #301 and #303, update the PackageMetadata protocol thus:

- Always provide a `Description` (even if found in the payload).
- Support multiline values for any field, even though the spec or build tools may still disallow it.
- For any multiline values, remove the line-continuation characters.
- Add a `json` property that converts the metadata into a JSON-compatible format per PEP 566.

Fixes #275, Fixes #276, Fixes #277.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,24 @@
+v4.0.0
+=======
+
+* #304: ``PackageMetadata`` as returned by ``metadata()``
+  and ``Distribution.metadata()`` now provides normalized
+  metadata honoring PEP 566:
+
+  - If a long description is provided in the payload of the
+    RFC 822 value, it can be retrieved as the ``Description``
+    field.
+  - Any multi-line values in the metadata will be returned as
+    such.
+  - For any multi-line values, line continuation characters
+    are removed. This backward-incompatible change means
+    that any projects relying on the RFC 822 line continuation
+    characters being present must be tolerant to them having
+    been removed.
+  - Add a ``json`` property that provides the metadata
+    converted to a JSON-compatible form per PEP 566.
+
+
 v3.10.0
 =======
 
--- a/importlib_metadata/__init__.py
+++ b/importlib_metadata/__init__.py
@@ -5,7 +5,6 @@ import csv
 import sys
 import zipp
 import email
-import string
 import pathlib
 import operator
 import textwrap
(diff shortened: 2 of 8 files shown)


jaraco:
Technically, this proposed change is a backward-incompatible one. Any projects that are depending on `Description` having line continuation indents will find that it doesn't now. This change actually manifest as a bug in pytest-checkdocs, fixed by jaraco/pytest-checkdocs@74d70d4fd6e021a0defe7e37f488458e320cf2af.

PR #313, opened 2021-05-23
Clarify DistributionFinder.Context.path type

pombredanne:
The DistributionFinder.Context.path property is a list of directory paths.

Signed-off-by: Philippe Ombredanne <pombredanne@nexb.com>


pombredanne:
I found myself scratching my head as to why passing a `Context(path='some/path')` object would not work out until I found out that this it was not a single `path` string (as the variable name was letting me believe) as but rather a list of path strings as in `sys.path`. I think this is worth clarifying. I could also provide a doc update along these lines.

PR #315, opened 2021-05-24
meta: add SimplePath protocol for PathDistribution

FFY00:
This is needed to properly annotate code that uses PathDistribution,
otherwise I need to keep redefining the protocol in my code.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,9 @@
+v4.1.0
+=======
+
+* #315: Add ``SimplePath`` protocol for interface clarity
+  in ``PathDistribution``.
+
 v4.0.1
 =======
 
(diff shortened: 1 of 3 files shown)


FFY00:
Please ping me when this goes to CPython, so that I can send a PR to typeshed.

jaraco:
I don't understand the diffcov failure. It says:

```
importlib_metadata/__init__.py (0.0%): Missing lines 26,786
```

It indicates that [line 26 is not covered](https://github.com/FFY00/importlib_metadata/blob/92084b80579be6b2bbbf1cd156ed8934313974e2/importlib_metadata/__init__.py#L26) and [line 786 is not covered](https://github.com/FFY00/importlib_metadata/blob/92084b80579be6b2bbbf1cd156ed8934313974e2/importlib_metadata/__init__.py#L786).

But those lines should be executed on import.

PR #317, opened 2021-05-27
Use normalized names to distinguish unique distributions for performance

### LINKED ISSUE #283: Improve efficiency of entry-point parsing
In #281, this project added support for uniqueness of distributions when parsing entry points. This change introduced some degradation in the performance when parsing entry points (due to need to load/inspect the metadata for every project). In that PR, a couple of suggestions were made to improve the performance:

- rely on the discovery name for the distribution rather than the proper name in metadata to disambiguate distributions
- ~~Rely on a custom, optimized parser instead of ConfigParser for parsing the entry points themselves.~~

Let's consider those two suggestions.

jaraco:
Fixes #283.

@anntzer Please have a look.

I found I couldn't use `path.stem` because `path` is a `SimplePath` and one implementation (`zipfile.Path`) doesn't implement `stem`.

I don't trust our benchmarks. Can you determine if this change has the performance benefit you were expecting?

### DIFF
--- a/importlib_metadata/__init__.py
+++ b/importlib_metadata/__init__.py
@@ -20,6 +20,7 @@ from ._compat import (
     NullFinder,
     PyPy_repr,
     install,
+    singledispatch,
 )
 from ._functools import method_cache
 from ._itertools import unique_everseen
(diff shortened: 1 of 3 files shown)


anntzer:
This works super nicely, and locally I get a huge (nearly 10x) speedup :tada:

Issue #300, opened 2021-03-30
importing `importlib_metadata` monkeypatches stdlib importlib breaking apis


asottile:
```python
import importlib.metadata
print(importlib.metadata.distribution('setuptools').entry_points[0])
__import__('importlib_metadata')
print(importlib.metadata.distribution('setuptools').entry_points[0])
```

```console
$ python3 t.py 
EntryPoint(name='alias', value='setuptools.command.alias:alias', group='distutils.commands')
Traceback (most recent call last):
  File "/tmp/y/venv/lib/python3.8/site-packages/importlib_metadata/__init__.py", line 224, in __getitem__
    return next(iter(self.select(name=name)))
StopIteration

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "t.py", line 4, in <module>
    print(importlib.metadata.distribution('setuptools').entry_points[0])
  File "/tmp/y/venv/lib/python3.8/site-packages/importlib_metadata/__init__.py", line 226, in __getitem__
    raise KeyError(name)
KeyError: 0
```

jaraco:
Looking at [the only code that's run implicitly on import](https://github.com/python/importlib_metadata/blob/b18a7eebaa3ae406d96bc566f016a1f91d36ec04/importlib_metadata/_compat.py#L18-L46), it acknowledges that the monkeypatching is sketchy and references #91 as the motivation. There's quite a bit of detail there, but it explains the challenge that this library faces in attempting to provide backward and forward compatibility for multiple versions of Python. In that issue, we settled on any import of importlib_metadata as giving it primacy for PathDistributions over the stdlib version.

More importantly, however, is that what you're encountering is an API incompatibility introduced in importlib_metadata 3.6. You can recreate the error in any version of `importlib_metadata>=3.6`.

```
~ $ pip-run -q 'importlib_metadata<3.6' setuptools -- -c "import importlib_metadata as im; im.distribution('setuptools').entry_points[0]"
~ $ pip-run -q 'importlib_metadata<3.7' setuptools -- -c "import importlib_metadata as im; im.distribution('setuptools').entry_points[0]"
Traceback (most recent call last):
  File "/var/folders/c6/v7hnmq453xb6p2dbz1gqc6rr0000gn/T/pip-run-o4vgk6e2/importlib_metadata/__init__.py", line 170, in __getitem__
    return next(iter(self.select(name=name)))
StopIteration

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
[...]

PR #323, opened 2021-05-30
EntryPoints compatibility

jaraco:
As described in #300, there are use-cases for the result of Distribution.entry_points that expect a mutable, list-like object with index access. These changes re-enable those behaviors while also publishing a Deprecation warning when mutable operations are enacted.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,22 @@
+v4.4.0
+=======
+
+* #300: Restore compatibility in the result from
+  ``Distribution.entry_points`` (``EntryPoints``) to honor
+  expectations in older implementations and issuing
+  deprecation warnings for these cases:
+
+  - ``EntryPoints`` objects are once again mutable, allowing
+    for ``sort()`` and other list-based mutation operations.
+    Avoid deprecation warnings by casting to a
+    mutable sequence (e.g.
+    ``list(dist.entry_points).sort()``).
+
+  - ``EntryPoints`` results once again allow
+    for access by index. To avoid deprecation warnings,
+    cast the result to a Sequence first
+    (e.g. ``tuple(dist.entry_points)[0]``).
+
 v4.3.1
 =======
 
--- a/importlib_metadata/__init__.py
+++ b/importlib_metadata/__init__.py
@@ -220,6 +220,14 @@ class EntryPoints(tuple):
         """
         Get the EntryPoint in self matching name.
         """
+        if isinstance(name, int):
+            warnings.warn(
+                "Accessing entry points by index is deprecated. "
+                "Cast to tuple if needed.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+            return super().__getitem__(name)
         try:
             return next(iter(self.select(name=name)))
         except StopIteration:
--- a/importlib_metadata/__init__.py
+++ b/importlib_metadata/__init__.py
@@ -209,7 +209,100 @@ class EntryPoint(
         return all(map(operator.eq, params.values(), attrs))
 
 
-class EntryPoints(tuple):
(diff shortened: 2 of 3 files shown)


jaraco:
@asottile I believe this change addresses the compatibility concerns raised in #300 and associated conversations. Would you have a look and let me know if it satisfies the use-cases you've encountered?

asottile:
it doesn't, this only reintroduces *deprecated* apis with no replacement

jaraco:
I'm sorry. I thought the replacement was pretty obvious from the deprecation warnings. Here's how one can adapt to the future behavior for index access:

```shell
importlib_metadata bugfix/300-entry-points-by-index $ .tox/python/bin/python -Wd -c "import importlib_metadata as md; print(md.distribution('pip').entry_points[0])"
EntryPoint(name='pip', value='pip._internal.cli.main:main', group='console_scripts')
<string>:1: DeprecationWarning: Accessing entry points by index is deprecated. Cast to tuple if needed.
importlib_metadata bugfix/300-entry-points-by-index $ .tox/python/bin/python -Wd -c "import importlib_metadata as md; print(tuple(md.distribution('pip').entry_points)[0])"
EntryPoint(name='pip', value='pip._internal.cli.main:main', group='console_scripts')
```

And for sorting:

```
importlib_metadata bugfix/300-entry-points-by-index $ .tox/python/bin/python -Wd -c "import importlib_metadata as md; eps = md.distribution('pip').entry_points; eps.sort(); print(eps[0].name)"
pip
<string>:1: DeprecationWarning: EntryPoints list interface is deprecated. Cast to list if needed.
<string>:1: DeprecationWarning: Accessing entry points by index is deprecated. Cast to tuple if needed.
importlib_metadata bugfix/300-entry-points-by-index $ .tox/python/bin/python -Wd -c "import importlib_metadata as md; eps = list(md.distribution('pip').entry_points); eps.sort(); print(eps[0].name)"
pip
# even better - prefer 'sorted' to sort in place:
importlib_metadata bugfix/300-entry-points-by-index $ .tox/python/bin/python -Wd -c "import importlib_metadata as md; eps = sorted(md.distribution('pip').entry_points); print(eps[0].name)"
pip
```

Historically, I've put replacements like this in the changelog. Would that suffice?

asottile:
both of those are significantly worse than the apis being replaced, O(1) -> O(n) copy, in-place sort to O(N) space

jaraco:
Acknowledged, the performance is sub-optimal. In my experience, the cost of O(n) operations is often negligible for n < 10000. I'm unsure if the added cost here is a practical concern or not, but I suspect it's a bearable cost. I do wish to reduce the dependence on mutable types, but leave open the possibility of a downstream consumer owning that behavior if they wish.

I'd ultimately like to make `entry_points()` return a generator to minimize the immediate cost of the call and to enable efficient use-cases like the one you describe and perhaps others. I chose not to make that optimization in this round due as I was unaware of any use-cases that demanded it and the implementation details seemed non-trivial given other goals (simple re-usability of the result).

For now, I propose to either (a) accept the performance penalty where practical, or (b) devise alternative approaches for these specialized use cases outside of the default interface.

It may make sense for `EntryPoints` to present a `group_map` or similar method that provides an O(1) lookup by group. I've seen one use case in `virtualenv` that may benefit from this optimization. Another approach may be to allow `entry_points` to accept a `container` parameter that would allow any caller to develop a specialized container for the iterable of `EntryPoint` objects. I welcome bug reports to describe the use-cases that would benefit from such a feature.

asottile:
the performance regression here is unacceptable.  it causes code to be 10-100x slower

please actually consider feedback before paving over it -- you're not being empathetic to my asks

PR #325, opened 2021-06-18
bpo-44451: Reset DeprecationWarning filters in test_api.test_entry_points_by_index

hroncok:
This avoids the following error in CPython tests if DeprecationWarnings are ignored.

    ======================================================================
    ERROR: test_entry_points_by_index (test.test_importlib.test_metadata_api.APITests)
    Prior versions of Distribution.entry_points would return a
    ----------------------------------------------------------------------
      File "/builddir/build/BUILD/Python-3.10.0b3/Lib/test/test_importlib/test_metadata_api.py", line 145, in test_entry_points_by_index
        expected = next(iter(caught))
    StopIteration
    ----------------------------------------------------------------------
    Ran 1402 tests in 2.125s
    FAILED (errors=1, skipped=18, expected failures=1)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,10 @@
+v4.3.1
+=======
+
+* #320: Fix issue where normalized name for eggs was
+  incorrectly solicited, leading to metadata being
+  unavailable for eggs.
+
 v4.3.0
 =======
 
--- a/README.rst
+++ b/README.rst
@@ -40,7 +40,7 @@ were contributed to different versions in the standard library:
 
    * - importlib_metadata
      - stdlib
-   * - 3.10
+   * - 4.2
      - 3.10
    * - 1.4
      - 3.8
--- a/README.rst
+++ b/README.rst
@@ -40,7 +40,7 @@ were contributed to different versions in the standard library:
 
    * - importlib_metadata
      - stdlib
-   * - 4.2
+   * - 4.4
      - 3.10
    * - 1.4
      - 3.8
--- a/conftest.py
+++ b/conftest.py
@@ -1,4 +1,25 @@
+import sys
+
+
 collect_ignore = [
     # this module fails mypy tests because 'setup.py' matches './setup.py'
     'prepare/example/setup.py',
 ]
+
+
+def pytest_configure():
+    remove_importlib_metadata()
+
+
+def remove_importlib_metadata():
+    """
+    Because pytest imports importlib_metadata, the coverage
+    reports are broken (#322). So work around the issue by
+    undoing the changes made by pytest's import of
+    importlib_metadata (if any).
+    """
+    if sys.meta_path[-1].__class__.__name__ == 'MetadataPathFinder':
+        del sys.meta_path[-1]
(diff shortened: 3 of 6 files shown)


jaraco:
Thanks for bringing this report and PR here. As I investigated and thought about it some more, I realized there was a deviation in how the deprecation warnings were being filtered in CPython versus here, a variance I introduced in 05e3288ea3413f7ad9317b14a266c024cb6a2458 (cpython branch only).

There are a couple of ways I imagine we can handle this. One could be to repeat the previous approach and apply this change to the cpython branch only. Another could be to apply both changes here in the main branch, but that would lead to unneeded and unused code (that could be later removed). A third approach I considered is to extract a function that captures the scope and purpose of this variance, something like:

```diff
diff --git a/tests/test_api.py b/tests/test_api.py
index 544ab7f4af4..0f51d6cb8d9 100644
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -3,6 +3,7 @@ import textwrap
 import unittest
 import warnings
 import importlib
+import contextlib
 
 from . import fixtures
 from importlib_metadata import (
@@ -17,6 +18,22 @@ from importlib_metadata import (
 )
 
 
+@contextlib.contextmanager
+def catch_warnings():
+    """
+    Catch and record warnings in a context manager.
+
+    When on CPython, ensure that DeprecationWarnings manifest
+    using default settings (not suppressed).
+
+    Ref bpo-44451.
+    """
+    with warnings.catch_warnings(record=True) as caught:
+        if __name__.startswith('test.test_importlib.'):
+            warnings.filterwarnings("default", category=DeprecationWarning)
+        yield caught
+
+
 class APITests(
     fixtures.EggInfoPkg,
     fixtures.DistInfoPkg,
@@ -122,7 +139,7 @@ class APITests(
         allowed casting those lists into maps by name using ``dict()``.
         Capture this now deprecated use-case.
         """
-        with warnings.catch_warnings(record=True) as caught:
+        with catch_warnings() as caught:
             eps = dict(entry_points(group='entries'))
 
         assert 'main' in eps
@@ -141,8 +158,7 @@ class APITests(
         See python/importlib_metadata#300 and bpo-44246.
         """
         eps = distribution('distinfo-pkg').entry_points
-        with warnings.catch_warnings(record=True) as caught:
-            warnings.filterwarnings("default", category=DeprecationWarning)
+        with catch_warnings() as caught:
             eps[0]
 
         # check warning
```

This approach would be easier to extend to other cases where it's necessary. I'm not sure it's worth the mess/hassle. I think I'm leaning toward option 1.

jaraco:
Weird that Github shows a merge conflict:

image

Because I'm able to merge the change using the Git client without any concern. I suspect something about the Github git client doesn't understand the rename.

```
importlib_metadata cpython $ git merge bpo-44451
Auto-merging Lib/test/test_importlib/test_metadata_api.py
Merge made by the 'recursive' strategy.
 Lib/test/test_importlib/test_metadata_api.py | 1 +
 1 file changed, 1 insertion(+)
```

So I'll push that merge commit and that will close this PR.

PR #335, opened 2021-08-05
fix(typing): Path-likes use truediv

### LINKED ISSUE #334: SimplePath protocol should use `__truediv__`, not `__div__`
`pathlib.Path` and `zip.Path` are implemented using `__truediv__`; `__div__` is not invoked for `/` in Python 3.

henryiii:
Fixes #334.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,9 @@
+v4.6.4
+======
+
+* #334: Correct ``SimplePath`` protocol to match ``pathlib``
+  protocol for ``__truediv``.
+
 v4.6.3
 ======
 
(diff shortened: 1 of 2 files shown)


henryiii:
If you'd like a test, something like this could be added:

```python
if typing.TYPE_CHECKING:
  _: SimplePath = typing.cast(Path, None)
```

FFY00 (approved):

FFY00:
@jaraco if you could you merge this and #336 and then make a release? Thanks!

jaraco:
In d491aea2da5 and 9e42358ef38, I attempted to add the test, but it doesn't fail, so I'll plan to merge without a regression test.

PR #336, opened 2021-08-05
export SimplePath

FFY00:
mypy is not happy about us using an attribute that isn't explicitedly
exported.

### DIFF
--- a/importlib_metadata/__init__.py
+++ b/importlib_metadata/__init__.py
@@ -38,6 +38,7 @@ __all__ = [
     'DistributionFinder',
     'PackageMetadata',
     'PackageNotFoundError',
+    'SimplePath',
     'distribution',
     'distributions',
     'entry_points',


jaraco:
How is it that mypy isn't happy? Oh, I notice the tests aren't running with all the plugins (notably mypy is missing) - at least when I run tests on Python 3.10.

jaraco:
I can see that [at least for this run](https://github.com/python/importlib_metadata/runs/3311781681?check_suite_focus=true#step:5:16), tests are run with mypy, but there's no failure. Can you demonstrate how the issue manifest for you?

layday:
> How is it that mypy isn't happy?

In strict mode mypy does not allow you to import symbols which aren't explicitly exported, either with `__all__` or with an `x as x` import.

jaraco:
> > How is it that mypy isn't happy?
> 
> In strict mode mypy does not allow you to import symbols which aren't explicitly exported, either with `__all__` or with an `x as x` import.

So is it no longer possible to import a symbol for internal use only (not for export)? I was expecting that `SimplePath` was not meant to be part of the public interface.

layday:
You can import it internally - @FFY00 was trying to import it in pypa/build, but that should no longer be necessary.

jaraco:
Cool. Thanks.

PR #338, opened 2021-08-13
Implement EntryPoint._asdict

jaraco:
In #337, Ronny reports a missed expectation - `EntryPoint` objects are constructed as `namedtuple` but fail to implement the `namedtuple._asdict` interface.

This change fulfills that expectation.


jaraco:
Superseded by #339

PR #354, opened 2021-10-30
metadata: remove Distribution._local

FFY00:
Importing an external module is anti-pattern and very unexpected
behavior. Furthermore, the specific builder implementation we are using
will provision an isolated virtual environmnent and perform the build
there, which is unwanted in various scenarious.

Perhaps there was a time this helper was needed, but we can now
remove this in favor of build.util.project_wheel_metadata[1].

[1] https://github.com/pypa/build/blob/82051d509a87124a46f3766134c11fc8aee9b86a/src/build/util.py#L27

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,13 @@
+v4.9.0
+======
+
+* #354: Removed ``Distribution._local`` factory. This
+  functionality was created as a demonstration of the
+  possible implementation. Now, the
+  `pep517 <https://pypi.org/project/pep517>`_ package
+  provides this functionality directly through
+  `pep517.meta.load <https://github.com/pypa/pep517/blob/a942316305395f8f757f210e2b16f738af73f8b8/pep517/meta.py#L63-L73>`_.
+
 v4.8.1
 ======
 
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -10,14 +10,12 @@ versions of Python), this can eliminate the need to use the older and less
 efficient ``pkg_resources`` package.
 
 ``importlib_metadata`` supplies a backport of
-:doc:`importlib.metadata <library/importlib.metadata>` as found in
-Python 3.8 and later for earlier Python releases.  Users of
-Python 3.8 and beyond are encouraged to use the standard library module
-when possible and fall back to ``importlib_metadata`` when necessary.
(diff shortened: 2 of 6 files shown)


jaraco:
Unfortunately, it doesn't look like `build.util.project_wheel_metadata` obviates this behavior. For example, the 'metadata' returned only includes the field-based metadata. It doesn't include other files in the metadata such as entry points. It's a lot more difficult to construct the full metadata because it requires multiple files, and to achieve that, pep517 constructs an in-memory zipfile.

Also a bit of a nitpick, but I don't see how "wheel" is relevant to all of these functions. A wheel isn't generated; only the project metadata. I'd like to see a function that doesn't involve irrelevant words like `util` and `wheel` to get the metadata.

FFY00:
> Unfortunately, it doesn't look like `build.util.project_wheel_metadata` obviates this behavior. For example, the 'metadata' returned only includes the field-based metadata. It doesn't include other files in the metadata such as entry points. It's a lot more difficult to construct the full metadata because it requires multiple files, and to achieve that, pep517 constructs an in-memory zipfile.

We can introduce further helpers to make that possible.

> Also a bit of a nitpick, but I don't see how "wheel" is relevant to all of these functions. A wheel isn't generated; only the project metadata. I'd like to see a function that doesn't involve irrelevant words like `util` and `wheel` to get the metadata.

It is the metadata that is generated for the wheel. The name comes from PEP 517.

https://www.python.org/dev/peps/pep-0517/#prepare-metadata-for-build-wheel

A wheel is actually generated as a fallback if the backend does not implement the metadata hook, as recommended.

sdists also contain metadata, which may be different from the wheel, and there is no guarantee that further binary formats that may be introduced in the future must have the same metadata as the wheel.
One way I can imagine this happening is introducing a new format that also takes care of binary dependencies for eg., where we would most likely have different metadata under certain scenarios.

I know I am being a bit conservative about the naming and origin of the data, but I didn't feel like it containing "wheel" in the name really hurt anything. Maybe I am an outlier, but my first questing after seeing a function named "project_metadata" would be "which metadata? from where?", as there are multiple metadata sources.

jaraco:
Oh! I see that pep517 [already provides this same functionality](https://github.com/pypa/pep517/blob/a942316305395f8f757f210e2b16f738af73f8b8/pep517/meta.py#L63-L73). Let's get rid of it here.

PR #355, opened 2021-11-09
fix changelog

benoit-pierre:


jaraco:
Since I released the same change in both 2.1.2 and 4.8.2, I meant for them to share a changelog entry. I _could_ duplicate the changelog entry, but it seems fine to just let the 4.8.2 appear as a null change against the change in 2.1.2. You'll see that other entries in the changelog are in chronological order and not in version order. So for consistency and to ensure that 2.1.2 has a changelog, I believe the current format is correct.

benoit-pierre:
Then merge the entries `v4.8.2 / v2.1.2`, or better yet keep the v2.x.x changelog to the v2 branch? Having both in the main branch seem highly illogical, IHMO.

benoit-pierre:
It would be great if each Github release contained the relevant changelog too.

PR #358, opened 2021-12-15
Fix requirement produced for git URL in extra

### LINKED ISSUE #357: InvalidRequirement produced for git URL in extra
Originally filed at https://github.com/pypa/packaging/issues/494

```
mkdir reqparse
cd reqparse

cat > setup.py <<HERE
from setuptools import setup

setup(
    name="reqparse",
    extras_require={
        "test": [
            "gym-retro @ git+https://github.com/openai/retro.git@v0.8.0",
        ]
    },
)
HERE

python3 -m venv env
source env/bin/activate
python3 -m pip install packaging -e .

cat reqparse.egg-info/requires.txt

# this now gives me
# packaging.requirements.InvalidRequirement: Parse error at "'extra =='": Expected string_end
python3 -c '
import importlib.metadata
from packaging.requirements import Requirement
Requirement(importlib.metadata.distribution("reqparse").requires[0])
'
```

This results in the equivalent of:
```
from packaging.requirements import Requirement
Requirement('gym-retro@ git+https://github.com/openai/retro.git@v0.8.0; extra == "test"')
```

uranusjr suggested adding a space before the semicolon here: https://github.com/python/cpython/blob/908fd691f96403a3c30d85c17dd74ed1f26a60fd/Lib/importlib/metadata/__init__.py#L678

Happy to make a PR for that — if so, should I also make a PR to CPython, or is there a separate process for porting?

hauntsaninja:
Fixes #357 , as suggested by @uranusjr

### DIFF
--- a/importlib_metadata/__init__.py
+++ b/importlib_metadata/__init__.py
@@ -702,7 +702,7 @@ class Distribution:
             if extra and markers:
                 markers = f'({markers})'
             conditions = list(filter(None, [markers, make_condition(extra)]))
-            return '; ' + ' and '.join(conditions) if conditions else ''
+            return ' ; ' + ' and '.join(conditions) if conditions else ''
 
         for section in sections:
             yield section.value + parse_condition(section.name)
--- a/importlib_metadata/__init__.py
+++ b/importlib_metadata/__init__.py
@@ -696,16 +696,25 @@ class Distribution:
         def make_condition(name):
             return name and f'extra == "{name}"'
 
-        def parse_condition(section):
+        def quoted_marker(section):
             section = section or ''
             extra, sep, markers = section.partition(':')
             if extra and markers:
                 markers = f'({markers})'
             conditions = list(filter(None, [markers, make_condition(extra)]))
-            return ' ; ' + ' and '.join(conditions) if conditions else ''
+            return '; ' + ' and '.join(conditions) if conditions else ''
+
+        def url_req_space(req):
+            """
+            PEP 508 requires a space between the url_spec and the quoted_marker.
+            Ref python/importlib_metadata#357.
+            """
+            # '@' is uniquely indicative of a url_req.
+            return ' ' * ('@' in req)
 
         for section in sections:
-            yield section.value + parse_condition(section.name)
+            space = url_req_space(section.value)
+            yield section.value + space + quoted_marker(section.name)
 
 
 class DistributionFinder(MetaPathFinder):
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -245,20 +245,20 @@ class APITests(
 
             [extra1]
             dep4
-            dep5@ git+https://example.com/python/dep.git@v1.0.0
+            dep6@ git+https://example.com/python/dep.git@v1.0.0
 
             [extra2:python_version < "3"]
-            dep6
+            dep5
             """
         )
         deps = sorted(Distribution._deps_from_requires_text(requires))
         expected = [
             'dep1',
             'dep2',
-            'dep3 ; python_version < "3"',
-            'dep4 ; extra == "extra1"',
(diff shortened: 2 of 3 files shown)


jaraco:
I cherry-picked the tests onto main and verified that they do indeed fail without the patch:

```
_______________________________________________ APITests.test_more_complex_deps_requires_text _______________________________________________

self = <tests.test_api.APITests testMethod=test_more_complex_deps_requires_text>

    def test_more_complex_deps_requires_text(self):
        requires = textwrap.dedent(
            """
            dep1
            dep2
    
            [:python_version < "3"]
            dep3
    
            [extra1]
            dep4
            dep5@ git+https://example.com/python/dep.git@v1.0.0
    
            [extra2:python_version < "3"]
            dep6
            """
        )
        deps = sorted(Distribution._deps_from_requires_text(requires))
        expected = [
            'dep1',
            'dep2',
            'dep3 ; python_version < "3"',
            'dep4 ; extra == "extra1"',
            'dep5@ git+https://example.com/python/dep.git@v1.0.0 ; extra == "extra1"',
            'dep6 ; (python_version < "3") and extra == "extra2"',
        ]
        # It's important that the environment marker expression be
        # wrapped in parentheses to avoid the following 'and' binding more
        # tightly than some other part of the environment expression.
    
>       assert deps == expected
E       AssertionError: assert ['dep1', 'dep... == "extra2"'] == ['dep1', 'dep... == "extra2"']
E         At index 2 diff: 'dep3; python_version < "3"' != 'dep3 ; python_version < "3"'
E         Use -v to get the full diff

tests/test_api.py:267: AssertionError
```

I do think since the failure doesn't specifically call out why it failed, there should be a comment similar to the "It's important..." comment that's already there, explaining what we learned in #357.

jaraco:
The tests are failing in Python 3.6 due to mypy complaining about no `importlib.resources` module. That test wasn't failing before, so I suspect something has changed in mypy.

tests/test_api.py
```diff
@@ -245,18 +245,20 @@ def test_more_complex_deps_requires_text(self):
...
             """
         )
         deps = sorted(Distribution._deps_from_requires_text(requires))
         expected = [
             'dep1',
             'dep2',
-            'dep3; python_version < "3"',
-            'dep4; extra == "extra1"',
-            'dep5; (python_version < "3") and extra == "extra2"',
+            'dep3 ; python_version < "3"',
```
jaraco, line 258: Reading the spec, the name_req doesn't require the space, so I'm slightly inclined not to add it because it's allowed. I'd like to explore what it would take to add the space only where it's required (provide the minimally-satisfying syntax).

jaraco: Fixed in d8aaa72.

jaraco:
> I do think since the failure doesn't specifically call out why it failed, there should be a comment similar to the "It's important..." comment that's already there, explaining what we learned in #357.

I no longer thing it's important to call out the why in the tests because the implementation calls out the rationale.

importlib_metadata/__init__.py
```diff
@@ -696,16 +696,25 @@ def _convert_egg_info_reqs_to_simple_reqs(sections):
...
             conditions = list(filter(None, [markers, make_condition(extra)]))
             return '; ' + ' and '.join(conditions) if conditions else ''
 
+        def url_req_space(req):
+            """
+            PEP 508 requires a space between the url_spec and the quoted_marker.
+            Ref python/importlib_metadata#357.
+            """
+            # '@' is uniquely indicative of a url_req.
+            return ' ' * ('@' in req)
```
uranusjr, line 713: _Technically_ this is not correct because `@` can be found in a version here; the version format in an `egg-info` is not PEP 508 compliant (it predates the standard). But adding a space is always fine, so no bugs will come out of this.

PR #362, opened 2022-01-14
ensure the arg to re.sub is a string object

balducci:
hi
for me, this fixes build failures of firefox, gimp and matplotlib with the error message:

```expected string or bytes-like object```


jaraco:
While this patch might fix the proximate error, I'd like to see a deeper understanding of the cause. Why is name not a string or bytes-like object?

If this failure does indeed indicate a failure, I'd also like to see a regression test capturing the cause.

jaraco:
Closing as it's growing stale. Feel free to re-engage on the matter.

PR #370, opened 2022-02-26
Add support for resolving dependencies (declared and missing) for an EntryPoint.

### LINKED ISSUE #368: differences in behavior with `pkg_resources`' `Entrypoint.load`
`importlib_metadata` implementation of `Entrypoint.load` does not check the requirements of the entrypoint's distribution, unlike `pkg_resources` (which raises `DistributionNotFound` if a requirement is missing, or `ContextualVersionConflict` in case of conflict). If this is expected, then that difference should at least be mentioned in the documentation, since `importlib_metadata` presents itself as a replacement for `pkg_resources`' API.

jaraco:
- Add basic degenerate tests for EntryPoint.deps.
- Add EntryPoint.deps for resolving all and missing deps for any entry point. Fixes #368.
- packaging is required to parse extras.
- Add meaningful test for EntryPoints.deps.
- ⚫ Fade to black.
- ⚫ Fade to black.
- Ignore order in expected deps
- Skip relevant tests when packaging is unavailable.

### DIFF
--- a/importlib_metadata/__init__.py
+++ b/importlib_metadata/__init__.py
@@ -153,7 +153,7 @@ class DeprecatedTuple:
 
 class Deps(set):
     """
-    A set of packaging.requirement.Requirements.
+    An iterable of packaging.requirement.Requirements.
     """
 
     @property
@@ -175,6 +175,21 @@ class Deps(set):
             return False
         found_ver = Version(dist.version)
         return found_ver in req.specifier
+
+    @classmethod
+    def for_entry_point(cls, ep):
+        """
+        Return Deps for any extras associated with the entry point.
+
+        Requires ep.dist to be defined.
+        """
+        from packaging.requirements import Requirement
+        return cls(
+            req
+            for req in map(Requirement, always_iterable(ep.dist.requires))
+            for extra in ep.extras
+            if req.marker.evaluate(dict(extra=extra))
+        )
 
 
 class EntryPoint(DeprecatedTuple):
@@ -247,18 +262,7 @@ class EntryPoint(DeprecatedTuple):
 
     @property
     def deps(self):
-        """
-        Return an iterable of Requirements for any extras associated
-        with this entry point. Requires self.dist to be defined.
-        """
-        from packaging.requirements import Requirement
-
-        return Deps(
-            req
-            for req in map(Requirement, always_iterable(self.dist.requires))
-            for extra in self.extras
-            if req.marker.evaluate(dict(extra=extra))
-        )
+        return Deps.for_entry_point(self)
 
     def _for(self, dist):
         vars(self).update(dist=dist)
--- a/importlib_metadata/__init__.py
+++ b/importlib_metadata/__init__.py
@@ -153,7 +153,7 @@ class DeprecatedTuple:
 
 class Deps(set):
     """
-    A set of requirements as strings.
+    A set of packaging.requirement.Requirements.
     """
     @property
     def missing(self):
@@ -162,9 +162,7 @@ class Deps(set):
         current environment. Requires 'packaging' to be
         installed.
         """
-        from packaging.requirements import Requirement
-
-        return itertools.filterfalse(self._is_installed, map(Requirement, self))
+        return itertools.filterfalse(self._is_installed, self)
 
     @staticmethod
     def _is_installed(req):
@@ -174,7 +172,7 @@ class Deps(set):
             dist = distribution(req.name)
         except PackageNotFoundError:
             return False
-        found_ver = Version(dist.version())
+        found_ver = Version(dist.version)
         return found_ver in req.specifier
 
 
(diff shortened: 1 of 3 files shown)


jaraco:
@benoit-pierre: would you care to review?

benoit-pierre:
This only look at one level of dependencies, no?

importlib_metadata/__init__.py
```diff
@@ -151,6 +151,47 @@ def __getitem__(self, item):
...
+        Return Deps for any extras associated with the entry point.
+
+        Requires ep.dist to be defined.
+        """
+        from packaging.requirements import Requirement
+        return cls(
+            req
+            for req in map(Requirement, always_iterable(ep.dist.requires))
+            for extra in ep.extras
+            if req.marker.evaluate(dict(extra=extra))
```
benoit-pierre, line 191:
```suggestion
            for extra in [None] + ep.extras
            if  req.marker is None or req.marker.evaluate(dict(extra=extra))
```

jaraco:
> This only look at one level of dependencies, no?

Yes. That's right. It's assuming that if the direct dependencies are satisfied, then transitive dependencies should have been satisfied.

importlib_metadata/__init__.py
jaraco: It looks like there may be two untested use-cases you're attempting to address here.

1. req.marker could be None, so would trigger an AttributeError.
2. The null extra should be implied.

I think I agree with (1) and I'll try to expand the tests to capture that expectation.

For (2), I'm less sure - under what conditions would an entry point need the null extra? Can this project rely on the expectation that if a package is installed that its non-extra dependencies are installed?

jaraco: I've addressed (1) in 7e07b739ff84.

benoit-pierre: It's not just the case of a missing dependency, what about if the dependency is there, but not compatible?

IMHO, if the feature does not handle all those use cases (missing / conflicting non-extra dependencies, or transitive dependencies), then what's the point? Better to not support it at all and be up-front about it in the documentation.

But considering the fact that an installed package metadata does not track requested extras (which means that something like `pip check` can't possibly detect a number of issues), accurately checking for missing / conflicting requirements when loading an entry point is an important feature. Even if apparently PYPA has decided to forego any kind of sanity check in generated console scripts launcher...

jaraco: Acknowledged. Thanks for explaining.

I do believe what you're seeking is a feature that I'd like to consider separately - that of verifying that a Distribution and its transitive dependencies are satisfied.

This feature is meant primarily to determine which additional dependencies are required for a given entry point and to report any that aren't already satisfied, leaving it to another function to handle what to do about that condition (install the dependency, raise an error, log a warning, ...).

I'd like to leave it as an additional exercise/feature to perform checks.

But now that I think about it, perhaps your right and this feature adds little value. After all, whatever feature checks for the required dependencies of an entry point may still want to check that the dependencies are met for the distribution (with relevant extras).

Something like `iter_unmet_dependencies(dist, extras=[])`, which could be called with `ep.dist, ep.extras` to validate an entry point.

At this point, I'm convinced the proposed feature and the 'unmet dependencies' feature is undesirable for this package (mainly because 'packaging' is unavailable). I'll work on creating a separate library that provides this functionality.

jaraco: BTW, I did try running the tests with:

```diff
diff --git a/importlib_metadata/__init__.py b/importlib_metadata/__init__.py
index a9df4ccf4846..346f81c70149 100644
--- a/importlib_metadata/__init__.py
+++ b/importlib_metadata/__init__.py
@@ -188,7 +188,7 @@ class Deps(set):
         return cls(
             req
             for req in map(Requirement, always_iterable(ep.dist.requires))
-            for extra in ep.extras
+            for extra in [None] + ep.extras
             if req.marker and req.marker.evaluate(dict(extra=extra))
         )
 ```

And the tests still pass, indicating that the technique of a `None` extra doesn't match the null marker. It probably would need to include `if not extra or (req.marker and req.marker.evaluate(...))`.

benoit-pierre: I don't understand. If you also want the non-optional dependencies, then you need to change the test to `req.marker is None or req.marker.evaluate(dict(extra=extra))`. `not extra or (req.marker and req.marker.evaluate(...))` would mean that a requirement of the form `python-xlib; "linux" in sys_platform` is not ignored on non-Linux platforms, no?

jaraco:
In the [jaraco/nspektr project](/jaraco/nspektr), I've drafted a library that will traverse requirements in the environment and report missing dependencies for an entry point. Please take a look and let me know what you think.

importlib_metadata/__init__.py
jaraco: Yes, good point. In a subsequent implementation (nspektr), I separate the concerns of resolving extra dependencies and regular dependencies and use a NullMarker to resolve an object on which `.evaluate` is always run.

PR #373, opened 2022-03-10
remove references to deprecated sequence keyed by group interface

graingert:


docs/using.rst
```diff
@@ -79,7 +79,7 @@ Query all entry points::
     >>> eps = entry_points()
 
 The ``entry_points()`` function returns an ``EntryPoints`` object,
-a sequence of all ``EntryPoint`` objects with ``names`` and ``groups``
+a collection of all ``EntryPoint`` objects with ``names`` and ``groups``
```
graingert, line 82: the sequence interface is deprecated:

```
DeprecationWarning: Accessing entry points by index is deprecated. Cast to tuple if needed.
```

docs/using.rst
```diff
@@ -42,9 +42,9 @@ You can get the version string for ``wheel`` by running the following::
     >>> version('wheel')
     '0.32.3'
 
-You can also get the set of entry points keyed by group, such as
+You can also get a collection of entry points selectable by group, such as
```
graingert, line 45: it's actually selectable by any of the EntryPoint attributes

```suggestion
You can also get a collection of entry points selectable by name, group or value, such as
```

jaraco: I'd probably not want to emphasize by value, as I don't expect legitimate use-cases to care about select by value. By name on the other hand seems a likely usage. Maybe.

```suggestion
You can also get a collection of entry points selectable by group or name (or other attribute), such as
```

jaraco:
This is great. The terminology was imprecise, and `Collection` is the right type. Thanks for proposing.

docs/using.rst
graingert: Or other attribute is a bit odd when there's only one other attribute

jaraco: Other possible attributes include `dist`, `module`, `attr`, and `extras`. See dbe114cbdc49.
```suggestion
You can also get a collection of entry points selectable by properties of the EntryPoint (typically 'group' or 'name'), such as
```

PR #377, opened 2022-05-12
fix `PathDistribution._normalized_name` implementation

benoit-pierre:
- correctly handle paths with no embedded version (e.g. `pkg.egg-info` must yield the name `pkg`, not `pkg.egg`)
- apply PEP 503 normalization to the extracted names (e.g.: `zope..inter_face-4.2.dist-info` must yield the name `zope_inter_face`)

Both are necessary, or `entry_points(…)` can yield the entry-points of a shadowed distribution. For example: with a version of `mypkg` in the system' site-packages directory when working from another development checkout of the same package (with a `mypgk.egg-info` directory mishandled by the first bug).

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -14,6 +14,7 @@ from importlib_metadata import (
     EntryPoint,
     MetadataPathFinder,
     PackageNotFoundError,
+    distribution,
     distributions,
     entry_points,
     metadata,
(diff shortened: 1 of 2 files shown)


jaraco:
Thanks @benoit-pierre for this proposal. Unfortunately, I'm struggling a little bit with it, mainly because it fixes two unreported bugs in one commit. I'm also uneasy about the added complexity that "normalized names" is adding here (requiring fixtures with multiple return values). Would you mind writing up a bug report describing how you encountered this bug and how we might replicate it (and prove the fix)?

I'm a little disappointed in the test suite right now. It seems the performance tests aren't running. Also, the linter didn't catch that `name` was shadowed and never used, which might have been an indicator that something was wrong.

jaraco:
In #379, I've addressed the first issue.

tests/test_main.py
```diff
@@ -87,15 +88,17 @@ def pkg_with_dashes(site_dir):
...
     def test_dashes_in_dist_name_found_as_underscores(self):
         """
         For a package with a dash in the name, the dist-info metadata
         uses underscores in the name. Ensure the metadata loads.
         """
-        pkg_name = self.pkg_with_dashes(self.site_dir)
-        assert version(pkg_name) == '1.0'
+        pkg_name, norm_pkg_name = self.pkg_with_dashes(self.site_dir)
+        dist = distribution(pkg_name)
+        assert dist._normalized_name == norm_pkg_name
```
jaraco, line 100: Instead of testing the internal implementation details, I'd rather have a test that captures the failed expectation - namely that entry points of two packages whose file system names vary only by normalization won't deduplicate.

jaraco:
In #381, I've found I can capture the missed expectation by simply tweaking the test for `test_entry_points_unique_packages` to include the expected normalization.

jaraco:
These changes are superseded by #379 and #381.

PR #381, opened 2022-05-21
Fix expectation on unique packages to include normalization.

jaraco:


jaraco:
I would like to actually test the behavior of `unique(distributions())`, but that function is currently local to the `entry_points` function. I've wanted to expose that function, so I might do that now.

Issue #380, opened 2022-05-21
Performance tests aren't running


jaraco:
I notice the performance tests aren't running. Or at least they aren't reporting any details. I'll need to investigate and see if maybe a release of pytest-perf has broken the tests.

jaraco:
Fixed in pytest-perf 0.12.1.

PR #386, opened 2022-06-13
Improve directory scanning logic

rossburton:


importlib_metadata/__init__.py
```diff
@@ -831,7 +831,13 @@ def __init__(self, path: FastPath):
         self.infos = FreezableDefaultDict(list)
         self.eggs = FreezableDefaultDict(list)
 
-        for child in path.children():
+        for child in sorted(path.children()):
```
jaraco, line 834: The Lookup is part of FastPath, which is highly optimized. Throwing a sort in here could noticeably affect performance. What's the purpose of sorting? What bug does it fix?

importlib_metadata/__init__.py
```diff
@@ -831,7 +831,13 @@ def __init__(self, path: FastPath):
...
+            # Empty directories aren't interesting
+            childpath = path.joinpath(child)
+            with suppress(PermissionError):
+                if childpath.is_dir() and not any(childpath.iterdir()):
+                    continue
```
jaraco, line 839: If needed, I'd like to see this block implemented as a `filter` on `path.children()` above.

importlib_metadata/__init__.py
```diff
@@ -831,7 +831,13 @@ def __init__(self, path: FastPath):
         self.infos = FreezableDefaultDict(list)
         self.eggs = FreezableDefaultDict(list)
 
-        for child in path.children():
+        for child in sorted(path.children()):
+            # Empty directories aren't interesting
+            childpath = path.joinpath(child)
+            with suppress(PermissionError):
+                if childpath.is_dir() and not any(childpath.iterdir()):
```
jaraco, line 838: I'm fairly confident this check is going to slow down the behavior. I think it also changes the behavior, causing metadata directories not to be recognized.

jaraco (changes requested):
Thanks for the contrib.

I believe what you have here is a couple of changes to designed behavior (sorting and skipping of empty directories). Aside from the performance concerns (which aren't accurately measured currently due to #380), if we wish to change the designed behavior, we'll want to do a few things:

- Capture a bug describing what is inadequate about the current behavior and why different behavior would be beneficial.
- Describe a design that might fix the issue (this PR serves as a draft design).
- Write some tests to capture the next expectation.

You're welcome to reply here or open a bug to begin the process I've recommended above. Also, consider reading [How to write the perfect pull request](https://blog.jaraco.com/how-to-write-perfect-pull-request/).

jaraco:
I'm happy to revisit this after some of the concerns have been addressed. Closing for now.

PR #389, opened 2022-06-23
Add Distribution.from_name arg validation

jaraco:
Fixes python/cpython#93259

- Add xfail test capturing new expectation.
- Add a validator to Distribution.from_name

### DIFF
--- a/importlib_metadata/__init__.py
+++ b/importlib_metadata/__init__.py
@@ -29,6 +29,8 @@ from contextlib import suppress
 from importlib import import_module
 from importlib.abc import MetaPathFinder
 from itertools import starmap
+from pydantic import Field, validate_arguments
+from pydantic.typing import Annotated
 from typing import List, Mapping, Optional, Union
 
 
@@ -548,7 +550,8 @@ class Distribution:
         """
 
     @classmethod
-    def from_name(cls, name):
+    @validate_arguments
+    def from_name(cls, name: Annotated[str, Field(min_length=1)]):
         """Return the Distribution for the given package name.
 
         :param name: The name of the distribution package to search for.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,11 @@
 [build-system]
-requires = ["setuptools>=56", "setuptools_scm[toml]>=3.4.1"]
+requires = [
+	"setuptools>=56",
+	"setuptools_scm[toml]>=3.4.1",
+	# temporary workaround for dependency race in setuptools_scm
+	# https://github.com/python/importlib_metadata/pull/389#issuecomment-1164912426
+	'pydantic; python_version < "3.8"',
+]
 build-backend = "setuptools.build_meta"
 
 [tool.black]
--- a/setup.cfg
+++ b/setup.cfg
@@ -19,6 +19,7 @@ python_requires = >=3.7
 install_requires =
 	zipp>=0.5
 	typing-extensions>=3.6.4; python_version < "3.8"
+	pydantic
 
 [options.packages.find]
 exclude =
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -5,6 +5,7 @@ import shutil
 import pathlib
 import tempfile
 import textwrap
+import functools
 import contextlib
 
 from .py39compat import FS_NONASCII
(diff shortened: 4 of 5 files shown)


jaraco:
For some reason, CI is failing me. The error is about 'setuptools_scm' not being able to find a version, but this code hasn't changed much from main, where the tests are passing. Is it the branch name maybe?

importlib_metadata/__init__.py
```diff
@@ -29,6 +29,8 @@
 from importlib import import_module
 from importlib.abc import MetaPathFinder
 from itertools import starmap
+from pydantic import Field, validate_arguments
+from pydantic.typing import Annotated
```
FFY00, line 33: Why is this being imported from `pydantic.typing` instead of `typing`?

importlib_metadata/__init__.py
```diff
@@ -548,7 +550,8 @@ def locate_file(self, path):
         """
 
     @classmethod
-    def from_name(cls, name):
+    @validate_arguments
+    def from_name(cls, name: Annotated[str, Field(min_length=1)]):
```
FFY00, line 554: Why are we introducing a dependency here? Won't we have to port this to CPython?

Also, personally, I'm unconvinced in introducing a dependency for such a simple check :sweat_smile:

jaraco:
In #390, I made another change in a branch/PR and it passed, so there must be something about this change that's affecting the build.

importlib_metadata/__init__.py
jaraco: That's what the [docs recommended](https://github.com/python/importlib_metadata/actions/runs/2552038722).

FFY00: Okay. I didn't find anything about it, in fact, the snippets I saw from the docs were importing it from `typing_extensions`, that's why I was confused.

importlib_metadata/__init__.py
jaraco: Over a year ago, I presented at the Python Language Summit a [proposal to use annotations](https://github.com/jaraco/transformers/blob/main/Annotations%20as%20transforms.md) to provide validators and transformers. The recommendation then was for me to explore pydantic, so I'm doing that now. I'm not certain I want to add a dependency, but I also want to explore the space. If it works well, I'll want to expand its usage to other similar patterns. As for porting to CPython, it's my understanding that parts of pydantic are being ported to CPython. I'm not sure if these parts are or not, but if not, this and related use-cases can help inform a rationale why they should.

jaraco:
> In #390, I made another change in a branch/PR and it passed, so there must be something about this change that's affecting the build.

I think I know the issue. Because this change introduces a dependency, and because setuptools_scm is relying on importlib_metadata, and because the local code with the dependency is present when running `setup.py develop`, `setuptools_scm` is failing to use `importlib_metadata` as it's declared in its own dependencies. It warns about the issue, but then also fails to resolve entry points and thus the version.

importlib_metadata/__init__.py
jaraco: Oops. That was the wrong link (now corrected).

jaraco:
I observe this change introduces some performance costs:

```
exercises.py:cached distribution: 0:00:00.000062 (+0:00:00.000023, 59%)
exercises.py:discovery: 0:00:00.000060 (+0:00:00.000021, 54%)
exercises.py:entry_points(): 0:00:00.004130 (+0:00:00.000029, 1%)
exercises.py:entrypoint_regexp_perf: 0:00:00.000096 (+-1 day, 23:59:59.999999, -1%)
exercises.py:uncached distribution: 0:00:00.000486 (+0:00:00.000063, 15%)
```

Adds between 21 and 63µs to discovery and distribution resolution (both cached and uncached).

That seems acceptable.

importlib_metadata/__init__.py
jaraco: As much as I'd like to explore the use of pydantic, I'm now realizing that the cost of adding a dependency here is very high, not only because of the CPython integration but also because of the race condition with setuptools_scm. I'm going to abandon this approach in favor of a manual check in #391. I would eventually like to explore this concept again, but today is not the day.

PR #393, opened 2022-06-25
Make "pip install --no-binary" possible on Python 3.7

### LINKED ISSUE #392: `pip install --no-binary` fails on Python 3.7
When running `pip install --no-binary :all: importlib-metadata` or even `pip install --no-binary importlib-metadata importlib-metadata` on Python 3.7, building fails due to a recursion error. That's because the latest `setuptools_scm` version introduced a dependency on `importlib-metadata` for `python_version < "3.8"`.

Maybe, `build-system requires` should be changed to require an older `setuptools_scm` version for `python_version < "3.8"`?

radoering:
Resolves #392

Avoid a recursion error when running `pip install --no-binary` for `python_version < "3.8"`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,9 @@
 [build-system]
-requires = ["setuptools>=56", "setuptools_scm[toml]>=3.4.1"]
+requires = [
+    "setuptools>=56",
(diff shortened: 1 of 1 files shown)


jaraco:
Thanks for the contrib, but I don't believe this is the right fix. If we decide otherwise in the issue, we can revive this PR.

Issue #396, opened 2022-07-14
AttributeError: 'PathDistribution' object has no attribute '_normalized_name'


jaraco:
When importlib_metadata introduced the name normalization in PathDistributions starting in v4.3, it introduced a subtle incompatibility in select environments.

In particular, if a provider uses an older importlib metadata (importlib_metadata < 4.3 or importlib.metadata on Python < 3.10), and in that provider they supply PathDistributions, calls to resolve entry points will fail when any distribution doesn't have a `_normalized_name` attribute.

This issue was reported in several issues including:
- pypa/setuptools#3102
- pypa/setuptools#3319
- pypa/setuptools#3452

Perhaps importlib_metadata should provide a compatibility shim to honor name normalization in distributions not supplying it (or at least not crash).

jaraco:
Fixed in #397 and scheduled for release in 4.13.0.

PR #397, opened 2022-07-18
Add compatibility for `PathDistributions` implemented with stdlib objects in Python 3.8/3.9

abravalheri:
This is an attempt to solve #396.

Changes:

- Added tests in 3.8/3.9 capturing expectations about custom `MetaPathFinder`s returning `PathDistribution` objects directly from stdlib.
- Added compatibility shim for `_normalized_name` during entry-point fetching
- Added compatibility shim from `EntryPoint.matches` during entry-point fetching

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,6 +16,11 @@ jobs:
         - ubuntu-latest
         - macos-latest
         - windows-latest
+        include:
+        - platform: ubuntu-latest
+          python: ~3.8.0-0
+        - platform: ubuntu-latest
+          python: ~3.9.0-0
     runs-on: ${{ matrix.platform }}
     steps:
       - uses: actions/checkout@v3
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,9 @@
+v4.13.0
+=======
+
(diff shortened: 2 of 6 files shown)


jaraco:
This looks good. I think I'd like to extract the compatibility shims into a module like `py39compat` or similar and make that behavior degenerate on Python 3.10+ so that when the future arrives, it's obvious what functionality needs to be removed... and so that functionality can be removed from the stdlib port right away. Similar for the tests. Let me know if that's something you wish to work on or something you'd like me to tackle.

abravalheri:
Thank you very much for the review Jason.

I tried to separate the files as you suggested[^1]. Please let me know if you had something different in mind.

[^1]: I had to use some tricks to avoid circular imports, please let me know if you have a different strategy that you would prefer.

jaraco:
>     1. I had to use some tricks to avoid circular imports, please let me know if you have a different strategy that you would prefer. [↩](#user-content-fnref-1-d69472787ca512b8674691a1b0baa186)

Oh, geez. I can see these issues must have been annoying. Thanks for tackling this concern - it will help future maintainers and ease integration into CPython.

PR #399, opened 2022-07-30
Clarify import vs. distribution packages in Using importlib_metadata docs

### LINKED ISSUE #398: Docs do not always clearly distinguish between import vs distribution packages
As [discussed on the Python Discourse](https://discuss.python.org/t/dynamic-versions-in-editable-installations/15220/44), the [Using importlib_metadata documentation](https://importlib-metadata.readthedocs.io/en/latest/using.html) and its [CPython equivalent](https://docs.python.org/3.12/library/importlib.metadata.html) do not always clearly distinguish between [import packages](https://packaging.python.org/en/latest/glossary/#term-Import-Package) and [distribution packages](https://packaging.python.org/en/latest/glossary/#term-Distribution-Package), leading to user confusion.

Therefore, these terms should be clarified, used consistently (avoiding the ambiguous unqualified "package") and linked to their PyPA glossary entries. Furthermore, an admonition specifically calling this out and suggesting users use `package_distributions()` to map between them would be helpful.

CAM-Gerlach:
As [requested on the Python Discourse](https://discuss.python.org/t/dynamic-versions-in-editable-installations/15220/46), addresses #398 (import vs. distribution packages not being clear) by:

* Using "distribution package"/"import package" on first use in each section
* Disambiguating the unqualified term "package" (preferring "distribution" and "import package", respectively) [^unsure]
* Linking the terms to the PyPA glossary once on first use in each section they are used in [^termlinks]
* Adding an admonition at the top explicitly calling out the differences and referring the reader to the `package_distributions()` function to map between them

Additionally, this incorporates a couple required/directly related changes:

* Fix a few grammar/textual problems on already-modified lines
* Add an intersphinx mapping to packaging.python.org to properly reference the terms and specs there [^intersphinx]
* Also uses such to reference the canonical, self-contained, up to date [Core metadata spec]() rather than an old, incremental, historical PEP. [^coremetadata]

Fix #398

Reviewer notes:

[^unsure]: I'm not 100% sure myself about the disambiguations in the highly technical ``Extending the search algorithm`` Explanation—please check to confirm I've gotten them correct.

[^termlinks]: So readers don't have to go hunting through other sections to find them, if they do so at all

[^intersphinx]:  If this needs to go in your skeleton, I can submit a PR there, drop that commit here and rebase accordingly once that is updated. Also, I was confused by some were written as dict literals, and some were added with `dict.update`, with no obvious or stated reason—I though it might have to do with the former being from your skeleton and the latter being local, but your skeleton looks identical, so I'm not sure what's going on or where packaging.python.org. Please advise and I'll be happy to add it in whichever manner is appropriate.

[^coremetadata]: If required, I can pull this out to a separate followup PR, though it is small enough I just added it here to avoid extra overhead on both our parts.

### DIFF
--- a/docs/using.rst
+++ b/docs/using.rst
@@ -4,21 +4,23 @@
  Using :mod:`!importlib_metadata`
 =================================
 
-``importlib_metadata`` is a library that provides access to installed
-package metadata, such as its entry points or its
-top-level name.  Built in part on Python's import system, this library
+``importlib_metadata`` is a library that provides access to
+the metadata of an installed distribution package, such as its entry points
+or its top-level import package name.
+Built in part on Python's import system, this library
 intends to replace similar functionality in the `entry point
 API`_ and `metadata API`_ of ``pkg_resources``.  Along with
 :mod:`importlib.resources`,
 this package can eliminate the need to use the older and less efficient
 ``pkg_resources`` package.
 
-By "installed package" we generally mean a third-party package installed into
-Python's ``site-packages`` directory via tools such as `pip
+By "installed package", we generally mean a third-party distribution package
+installed into Python's ``site-packages`` directory via tools such as `pip
 <https://pypi.org/project/pip/>`_.  Specifically,
-it means a package with either a discoverable ``dist-info`` or ``egg-info``
+it means a distribution with either a discoverable ``dist-info`` or ``egg-info``
 directory, and metadata defined by :pep:`566` or its older specifications.
-By default, package metadata can live on the file system or in zip archives on
+By default, distribution metadata can live on the file system
+or in zip archives on
 :data:`sys.path`.  Through an extension mechanism, the metadata can live almost
 anywhere.
 
@@ -33,7 +35,8 @@ anywhere.
 Overview
 ========
 
-Let's say you wanted to get the version string for a package you've installed
+Let's say you wanted to get the version string for a
+distribution package you've installed
 using ``pip``.  We start by creating a virtual environment and installing
 something into it::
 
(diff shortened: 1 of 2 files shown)


docs/using.rst
```diff
@@ -4,21 +4,39 @@
  Using :mod:`!importlib_metadata`
 =================================
 
-``importlib_metadata`` is a library that provides access to installed
-package metadata, such as its entry points or its
-top-level name.  Built in part on Python's import system, this library
+``importlib_metadata`` is a library that provides access to
+the metadata of an installed :term:`packaging:Distribution Package`,
+such as its entry points
+or its top-level :term:`packaging:Import Package` name.
```
jaraco, line 10: This phrasing is misleading, suggesting that a distribution package has exactly one import package, when a distribution package may have any number of package names (including none) and may also contain names of modules supplied by the distribution package. Maybe just "top-level names" would be more accurate.

jaraco (approved):
Looks good to me. Just one question.

docs/using.rst
jaraco:
```suggestion
or its top-level names (:term:`packaging:Import Package`s, modules, if any).
```

CAM-Gerlach:
Hey @jaraco sorry for missing this and thanks for taking care of it; I was extremely busy for the Artemis 1 launch attempts during this time and then must have accidentally dismissed it (I believe I mistakenly thought it was a different unrelated issue rather than this PR). You shouldn't have to, but in the future you're welcome to give me a ping on Discord, etc. just in case I miss something like this in the future, so you don't have to wait around for a month and then have to do it yourself. Thanks!

PR #401, opened 2022-07-30
Mention API ref & migration guide in importlib_metadata See Also

### LINKED ISSUE #400: Add note that API docs (and migration guide) can be found on importlib_metadata docs
As [discussed on the Python Discourse](https://discuss.python.org/t/dynamic-versions-in-editable-installations/15220/44), it could be made clearer in the [Using importlib_metadata documentation](https://importlib-metadata.readthedocs.io/en/latest/using.html), and thus its [CPython equivalent](https://docs.python.org/3.12/library/importlib.metadata.html), that the API documentation (and migration guide) can be found on the ``importlib_metadata`` docs site, as many users are not necessarily aware of such. 

Ideally, it would be included directly in the CPython docs either via Sphinx-Autodoc, manually or some other mechanism, given the linked version would be out of sync with what is actually included in the particular version of the CPython source, but this is a useful quick stopgap for now.

CAM-Gerlach:
Per #400 and as As [discussed on the Python Discourse](https://discuss.python.org/t/dynamic-versions-in-editable-installations/15220/44), this clarifies the  [Using importlib_metadata documentation](https://importlib-metadata.readthedocs.io/en/latest/using.html) (and thus its [CPython equivalent](https://docs.python.org/3.12/library/importlib.metadata.html)) that the API documentation (and migration guide) for the latest version of ``importlib.metadata` can be found on the ``importlib_metadata`` docs site.

Fix #400

### DIFF
--- a/docs/using.rst
+++ b/docs/using.rst
@@ -28,6 +28,12 @@ anywhere.
    https://importlib-metadata.readthedocs.io/
       The documentation for ``importlib_metadata``, which supplies a
       backport of ``importlib.metadata``.
+      This includes an `API reference
+      <https://importlib-metadata.readthedocs.io/en/latest/api.html>`__
+      for this module's classes and functions,
+      as well as a `migration guide
+      <https://importlib-metadata.readthedocs.io/en/latest/migration.html>`__
+      for existing users of ``pkg_resources``.
 
 
 Overview


docs/using.rst
```diff
@@ -28,6 +28,12 @@ anywhere.
    https://importlib-metadata.readthedocs.io/
       The documentation for ``importlib_metadata``, which supplies a
       backport of ``importlib.metadata``.
+      This includes an `API reference
+      <https://importlib-metadata.readthedocs.io/en/latest/api.html>`__
+      for this module's classes and functions,
+      as well as a `migration guide
+      <https://importlib-metadata.readthedocs.io/en/latest/migration.html>`__
+      for existing users of ``pkg_resources``.
```
jaraco, line 36: I'm a little concerned about adding multiple references and keeping those in sync. Why not just "..., which supplies a backport of ``importlib.metadata`` and includes an API reference and migration guide for users of pkg_resources." That would inform the users of the presence of those aspects without keeping them linked.

jaraco (approved):
LGTM. Just one question.

jaraco:
On further consideration, let's try the links. If they become onerous to maintain, we may replace them later with a less specific reference.

docs/using.rst
CAM-Gerlach: That's a good point. We could use `:ref:` via Intersphinx which would keep them in sync automatically, though that would require adding intersphinx to the CPython docs build config which I'm not sure the downstreams would like (since it requires either network access or a local cache when building docs). But if you think so, we can just go with it for now—you're the maintainer, of course.

PR #405, opened 2022-10-01
Remove compatibility shims for entry points.

jaraco:
- Remove SelectableGroups
- Remove DeprecatedList
- Remove compatibility for EntryPoints.__getitem__ by index.
- Remove support for cast of iterable of entry points to dict. Ref #97.
- Update changelog

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,9 @@
+v5.0.0
+======
+
+* #97, #284, #300: Removed compatibility shims for deprecated entry
+  point interfaces.
+
 v4.13.0
 =======
 
--- a/docs/using.rst
+++ b/docs/using.rst
@@ -164,11 +164,10 @@ for more information on entry points, their definition, and usage.
 The "selectable" entry points were introduced in ``importlib_metadata``
 3.6 and Python 3.10. Prior to those changes, ``entry_points`` accepted
 no parameters and always returned a dictionary of entry points, keyed
-by group. For compatibility, if no parameters are passed to entry_points,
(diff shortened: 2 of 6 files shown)


woutdenolf:
Starts to be rather cumbersome to support `python>=3.7`

```python
import sys

try:
    from importlib.metadata import entry_points
except ImportError:
    from importlib_metadata import entry_points

if sys.version_info >= (3,10):
    _entry_points = entry_points(group=namespace)
else:
    try:
        _entry_points = entry_points().get(namespace, [])
    except AttributeError:
        _entry_points = entry_points().select(namespace, [])
```

jaraco:
That's one way to do it. Other options include:

- Rely on importlib_metadata >= 3.6 for Python < 3.10. As a general rule, you should adopt `importlib_metadata` any time there's behavior from a future Python version but your application supports older Python versions.
- Use [backports.entry_points_selectable](https://pypi.org/project/backports.entry_points_selectable) (provides forward compatibility to Python 2.7+).
- Bypass `entry_points` altogether and re-implement its behavior using more fundamental interfaces (flake8 did this; I don't recommend it).

woutdenolf:
Thanks for the feedback!

> Rely on importlib_metadata >= 3.6 for Python < 3.10

Many projects just need `importlib.metadata` so they have `importlib-metadata; python_version < '3.8'`. Once python 3.7 gets dropped it should become simpler.

Python 3.10 still supports `entry_points().select(group=namespace)` so using `entry_points(group=namespace)` is not mandatory.

PR #408, opened 2022-10-03
Simplify compatibility wrappers

jaraco:
These changes limit the complexity of the compatibility wrappers and make it easier to reason about how to remove them when appropriate.

### DIFF
--- a/importlib_metadata/_py39compat.py
+++ b/importlib_metadata/_py39compat.py
@@ -1,7 +1,7 @@
 """
 Compatibility layer with Python 3.8/3.9
 """
-from typing import TYPE_CHECKING, Any, Optional, Tuple
+from typing import TYPE_CHECKING, Any, Optional
 
 if TYPE_CHECKING:  # pragma: no cover
     # Prevent circular imports on runtime.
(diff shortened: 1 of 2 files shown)


jaraco:
@abravalheri Would you mind reviewing this change? In particular, does the lack of replacing the EntryPoint object have compatibility concerns that you'd intentionally addressed?

importlib_metadata/_py39compat.py
```diff
@@ -19,30 +19,17 @@ def normalized_name(dist: Distribution) -> Optional[str]:
...
     try:
-        return ep, ep.matches(**params)
+        return ep.matches(**params)
     except AttributeError:
         from . import EntryPoint  # -> delay to prevent circular imports.
 
         # Reconstruct the EntryPoint object to make sure it is compatible.
-        _ep = EntryPoint(ep.name, ep.value, ep.group)
-        return _ep, _ep.matches(**params)
+        return EntryPoint(ep.name, ep.value, ep.group).matches(**params)
```
abravalheri, line 35: There is a small risk associated with re-using the `EntryPoint` objects as returned by an old `importlib.metadata` implementation:

- If someone uses the `importlib_metadata.entry_points()`/`importlib_metadata.EntryPoints.select()` API, they expect the result to be a collection of `importlib_metadata.EntryPoint` objects.
- However, with the change what they will get is a collection of `Union[importlib_metadata.EntryPoint, importlib.metadata.EntryPoint]` which is not strictly API compatible with their expectations (and therefore may cause runtime errors).

If this risk is considered small enough, the change should be relatively safe.

importlib_metadata/_py39compat.py
```diff
@@ -19,30 +19,17 @@ def normalized_name(dist: Distribution) -> Optional[str]:
     except AttributeError:
         from . import Prepared  # -> delay to prevent circular imports.
 
-        return Prepared.normalize(getattr(dist, "name", None) or dist.metadata['Name'])
+        return Prepared.normalize(dist.metadata['Name'])
```
abravalheri, line 22: In newer versions of `importlib_metadata`, `_normalized_name` is resolved by normalizing `self.name`. Without inheritance, this is equivalent to `dist.metadata['Name']`. Inheritance, however, allows users to overwrite it.

Is there any chance this impacts here? (If there is, probably the risk is very low anyway, because old implementations would not expose the `.name` property, and newer implementations will not fail when accessing `dist._normalized_name`)

abravalheri (approved):
Hi Jason, I added some minor comments explaining what are the risks I see with the change. They are probably very small risks so the change should be good to go.

Thank you very much for having a look on this.

importlib_metadata/_py39compat.py
jaraco: In this case, I agree the risk is low, so I'll proceed with this change.

importlib_metadata/_py39compat.py
jaraco: That's a really good point. I do believe there are implementations out there that override `.name`, so I'll back out this change.

PR #433, opened 2023-03-10
Fix formatting in test_api

FFY00:
Tests are failing on #432 because of this formatting issue, which is unrelated. I assume something in the formatter changed, causing the current code to be outdated, which is undesirable in a formatter :upside_down_face:


FFY00:
Now we're hitting #434

FFY00:
#435 is failing because of the lack of this change, so I am just gonna go ahead and merge this PR as it's so trivial.

PR #435, opened 2023-03-10
Sync PackageMetadata with email.message.Message

### LINKED ISSUE #434: mypy failing
mypy is failing in the tests with the following error

```
435: error: Incompatible return value type (got "Message", expected "PackageMetadata")  [return-value]
435: note: Following member(s) of "Message" have conflicts:
435: note:     Expected:
435: note:         def [_T] get_all(self, name: str, failobj: _T = ...) -> Union[List[Any], _T]
435: note:     Got:
435: note:         @overload
435: note:         def get_all(self, name: str, failobj: None = ...) -> Optional[List[Any]]
435: note:         @overload
435: note:         def [_T] get_all(self, name: str, failobj: _T) -> Union[List[Any], _T]
```

The typeshed definitions changed in https://github.com/python/typeshed/commit/b5a26d1ca258a86a6f133dfabf58790553dc21cf, causing `PackageMetadata` to fall out of sync. I think the right option here is just to bring `PackageMetadata` back in sync with `email.message.Message`

FFY00:
Fixes #434

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,10 @@
+v6.0.1
+======
+
+* #434: Expand protocol for ``PackageMetadata.get_all`` to match
+  the upstream implementation of ``email.message.Message.get_all``
+  in python/typeshed#9620.
+
 v6.0.0
 ======
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -61,9 +61,11 @@ intersphinx_mapping.update(
     ),
 )
 
-# Workaround for #316
 nitpick_ignore = [
+    # Workaround for #316
     ('py:class', 'importlib_metadata.EntryPoints'),
     ('py:class', 'importlib_metadata.SelectableGroups'),
     ('py:class', 'importlib_metadata._meta._T'),
+    # Workaround for #435
+    ('py:class', '_T'),
 ]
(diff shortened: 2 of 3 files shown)


jaraco:
This code makes me sad. It looks like the typing system requires overloads anywhere where a `None` is the default fallback object? It took me many minutes just to try to comprehend what's going on here and it's degrading the readability of the `get_all` method. The reader now has to incorporate additional concepts (`Optional`, `overload`) and reconcile the two interfaces just to reason about what, without typing checks, is a simple concept (`failobj=None`).

Moreover, this test fails the diffcov because it fails to cover the new line 23 (an ellipsis).

What ever happened to "elegance begets simplicity"?

Is there any expectation that this behavior will ever be simplified? I guess I can link to the PRs for those who need to track down the origin.

jaraco:
For reference - upstream change at https://github.com/python/typeshed/pull/9620 and possible long-term simplification at https://github.com/python/typing/issues/307.

jaraco:
This change also introduces errors in the docs build:

```
/home/runner/work/importlib_metadata/importlib_metadata/importlib_metadata/_meta.py:docstring of typing._overload_dummy:1: WARNING: py:class reference target not found: _T
```

PR #436, opened 2023-03-10
Temporarily disable 3.12 CI due to incompatibility in pyfakefs

FFY00:
3.12 is currently broken due to https://github.com/pytest-dev/pyfakefs/issues/770


jaraco:
No need to disable the tests. They're allowed to fail.

PR #437, opened 2023-03-11
Fix .files and inferred packages_distributions for .egg-info packages

### LINKED ISSUE #115: "files" is wrong when installed without "wheel"
In GitLab by [@blueyed](https://gitlab.com/blueyed) on Mar 17, 2020, 12:32

When "wheel" is not installed/used, installing a package will not contain "RECORD", and it falls back to reading "SOURCES.txt" (https://gitlab.com/python-devs/importlib_metadata/blob/3150ed4da9e1267d0787c6f4c1f8258a26a1dd93/importlib_metadata/__init__.py#L270), which then might result in paths not being `locate()`able, when a src-based setup is used:
```
setup(
    ...
    packages=find_packages('src'),
    package_dir={'': 'src'},
)
```

This is similar to https://gitlab.com/python-devs/importlib_metadata/-/issues/112, but in this case here it could use `installed-files.txt` from the egg-info.

jherland:
The first three commits fix #115 for the case where a `.egg-info` package includes `installed-files.txt`:
We prefer `installed-files.txt` over `SOURCES.txt` as a source for `Distribution.files`, as the former is deemed more accurate when it is available.

The last two commits attempts to fix #115 for the case where we have no choice but to parse `SOURCES.txt`:
This _assumes_ that `Distribution.files` should never contain a path to a non-existent file.
We simply remove non-existent files with a filter, thus removing any entries from `SOURCES.txt` that refer to files that were part of the source package, but not part of the final installation.

Commits:
- `tests/fixtures`: Fix `FilesDef` type to include `bytes` values
- Add tests for egg-info package with no installed modules
- `Distribution.files`: Prefer `*.egg-info/installed-files.txt` to `SOURCES.txt`
- Add tests for egg-info package with `.files` from inaccurate `SOURCES.txt`
- `Distribution.files`: Only return files that actually exist

### DIFF
--- a/importlib_metadata/__init__.py
+++ b/importlib_metadata/__init__.py
@@ -475,17 +475,18 @@ class Distribution(metaclass=abc.ABCMeta):
 
         @pass_none
         def make_files(lines):
-            return list(
-                filter(
-                    lambda package_path: package_path.locate().exists(),
-                    starmap(make_file, csv.reader(lines)),
-                )
-            )
+            return starmap(make_file, csv.reader(lines))
 
-        return make_files(
-            self._read_files_distinfo()
-            or self._read_files_egginfo_installed()
-            or self._read_files_egginfo_sources()
+        @pass_none
+        def skip_missing_files(package_paths):
+            return list(filter(lambda path: path.locate().exists(), package_paths))
+
+        return skip_missing_files(
+            make_files(
+                self._read_files_distinfo()
+                or self._read_files_egginfo_installed()
+                or self._read_files_egginfo_sources()
+            )
         )
 
     def _read_files_distinfo(self):
--- a/importlib_metadata/__init__.py
+++ b/importlib_metadata/__init__.py
@@ -478,7 +478,7 @@ class Distribution(metaclass=abc.ABCMeta):
             return list(
                 filter(
                     lambda package_path: package_path.locate().exists(),
-                    list(starmap(make_file, csv.reader(lines))),
+                    starmap(make_file, csv.reader(lines)),
                 )
             )
 
--- a/importlib_metadata/__init__.py
+++ b/importlib_metadata/__init__.py
@@ -512,7 +512,16 @@ class Distribution(metaclass=abc.ABCMeta):
         # But this subdir is only available in the PathDistribution's self._path
         # which is not easily accessible from this base class...
         subdir = getattr(self, '_path', None)
-        return text and subdir and [f'"{subdir}/{line}"' for line in text.splitlines()]
+        try:
+            if text and subdir:
+                ret = [
+                    str((subdir / line).resolve().relative_to(self.locate_file('')))
+                    for line in text.splitlines()
+                ]
+                return map('"{}"'.format, ret)
+        except Exception:
+            pass
+        return None
 
     def _read_files_egginfo_sources(self):
         """
--- a/importlib_metadata/__init__.py
+++ b/importlib_metadata/__init__.py
@@ -474,7 +474,12 @@ class Distribution(metaclass=abc.ABCMeta):
 
         @pass_none
         def make_files(lines):
-            return list(starmap(make_file, csv.reader(lines)))
+            return list(
+                filter(
+                    lambda package_path: package_path.locate().exists(),
+                    list(starmap(make_file, csv.reader(lines))),
+                )
+            )
 
         return make_files(
             self._read_files_distinfo()
--- a/importlib_metadata/__init__.py
+++ b/importlib_metadata/__init__.py
@@ -12,6 +12,7 @@ import warnings
 import functools
 import itertools
 import posixpath
+import contextlib
 import collections
 import inspect
 
@@ -513,16 +514,14 @@ class Distribution(metaclass=abc.ABCMeta):
         # But this subdir is only available in the PathDistribution's self._path
         # which is not easily accessible from this base class...
         subdir = getattr(self, '_path', None)
-        try:
-            if text and subdir:
-                ret = [
-                    str((subdir / line).resolve().relative_to(self.locate_file('')))
-                    for line in text.splitlines()
-                ]
-                return map('"{}"'.format, ret)
-        except Exception:
-            pass
-        return None
+        if not text or not subdir:
+            return
+        with contextlib.suppress(Exception):
+            ret = [
+                str((subdir / line).resolve().relative_to(self.locate_file('')))
+                for line in text.splitlines()
+            ]
+            return map('"{}"'.format, ret)
 
     def _read_files_egginfo_sources(self):
         """
--- a/importlib_metadata/__init__.py
+++ b/importlib_metadata/__init__.py
@@ -498,10 +498,10 @@ class Distribution(metaclass=abc.ABCMeta):
 
     def _read_files_egginfo_installed(self):
         """
-        installed-files.txt might contain literal commas, so wrap
-        each line in quotes. Also, the entries in installed-files.txt
-        are relative to the .egg-info/ subdir (not relative to the
-        parent site-packages directory that make_file() expects).
+        Read installed-files.txt and return lines in a similar
+        CSV-parsable format as RECORD: each file must be placed
+        relative to the site-packages directory, and must also be
+        quoted (since file names can contain literal commas).
 
         This file is written when the package is installed by pip,
         but it might not be written for other installation methods.
@@ -526,8 +526,9 @@ class Distribution(metaclass=abc.ABCMeta):
 
     def _read_files_egginfo_sources(self):
         """
-        SOURCES.txt might contain literal commas, so wrap each line
-        in quotes.
+        Read SOURCES.txt and return lines in a similar CSV-parsable
+        format as RECORD: each file name must be quoted (since it
+        might contain literal commas).
 
         Note that SOURCES.txt is not a reliable source for what
         files are installed by a package. This file is generated
--- a/importlib_metadata/__init__.py
+++ b/importlib_metadata/__init__.py
@@ -460,8 +460,8 @@ class Distribution(metaclass=abc.ABCMeta):
         :return: List of PackagePath for this distribution or None
 
         Result is `None` if the metadata file that enumerates files
-        (i.e. RECORD for dist-info or SOURCES.txt for egg-info) is
-        missing.
+        (i.e. RECORD for dist-info, or installed-files.txt or
+        SOURCES.txt for egg-info) is missing.
         Result may be empty if the metadata exists but is empty.
         """
 
@@ -476,7 +476,11 @@ class Distribution(metaclass=abc.ABCMeta):
         def make_files(lines):
             return list(starmap(make_file, csv.reader(lines)))
 
-        return make_files(self._read_files_distinfo() or self._read_files_egginfo())
+        return make_files(
+            self._read_files_distinfo()
+            or self._read_files_egginfo_installed()
+            or self._read_files_egginfo_sources()
+        )
 
     def _read_files_distinfo(self):
         """
@@ -485,10 +489,35 @@ class Distribution(metaclass=abc.ABCMeta):
         text = self.read_text('RECORD')
         return text and text.splitlines()
 
-    def _read_files_egginfo(self):
+    def _read_files_egginfo_installed(self):
+        """
+        installed-files.txt might contain literal commas, so wrap
+        each line in quotes. Also, the entries in installed-files.txt
+        are relative to the .egg-info/ subdir (not relative to the
+        parent site-packages directory that make_file() expects).
+
+        This file is written when the package is installed by pip,
+        but it might not be written for other installation methods.
+        Hence, even if we can assume that this file is accurate
+        when it exists, we cannot assume that it always exists.
+        """
+        text = self.read_text('installed-files.txt')
+        # We need to prepend the .egg-info/ subdir to the lines in this file.
+        # But this subdir is only available in the PathDistribution's self._path
+        # which is not easily accessible from this base class...
+        subdir = getattr(self, '_path', None)
+        return text and subdir and [f'"{subdir}/{line}"' for line in text.splitlines()]
+
+    def _read_files_egginfo_sources(self):
         """
         SOURCES.txt might contain literal commas, so wrap each line
         in quotes.
+
+        Note that SOURCES.txt is not a reliable source for what
+        files are installed by a package. This file is generated
+        for a source archive, and the files that are present
+        there (e.g. setup.py) may not correctly reflect the files
+        that are present after the package has been installed.
         """
         text = self.read_text('SOURCES.txt')
         return text and map('"{}"'.format, text.splitlines())
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -243,6 +243,32 @@ class EggInfoPkgPipInstalledNoModules(OnSysPath, SiteDir):
         build_files(EggInfoPkgPipInstalledNoModules.files, prefix=self.site_dir)
 
 
+class EggInfoPkgSourcesFallback(OnSysPath, SiteDir):
+    files: FilesDef = {
+        "starved_egg_pkg.egg-info": {
+            "PKG-INFO": "Name: starved_egg-pkg",
+            # SOURCES.txt is made from the source archive, and contains files
+            # (setup.py) that are not present after installation.
+            "SOURCES.txt": """
+                starved_egg_pkg.py
+                setup.py
+                starved_egg_pkg.egg-info/PKG-INFO
+                starved_egg_pkg.egg-info/SOURCES.txt
+            """,
+            # missing installed-files.txt (i.e. not installed by pip)
+            # missing top_level.txt
+        },
+        "starved_egg_pkg.py": """
+            def main():
+                print("hello world")
+            """,
+    }
+
+    def setUp(self):
+        super().setUp()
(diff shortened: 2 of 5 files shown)


FFY00 (approved):
Thanks @jherland!

This looks good to me, but I am unsure about the last commit (only returning files that exist on the file system), so I'd like @jaraco to also have a look if he's available.

jherland:
Thanks for the review @FFY00. I took a look at the failing tests. The 3.12 failures seem related to #436, and the other three are cases of #434/#435.

importlib_metadata/__init__.py
```diff
@@ -486,10 +495,35 @@ def _read_files_distinfo(self):
...
+        This file is written when the package is installed by pip,
+        but it might not be written for other installation methods.
+        Hence, even if we can assume that this file is accurate
+        when it exists, we cannot assume that it always exists.
+        """
+        text = self.read_text('installed-files.txt')
+        # We need to prepend the .egg-info/ subdir to the lines in this file.
+        # But this subdir is only available in the PathDistribution's self._path
+        # which is not easily accessible from this base class...
+        subdir = getattr(self, '_path', None)
```
jaraco, line 514: This approach feels a little hacky. Perhaps it would make sense instead to prescribe an interface for `Distribution` objects to resolve paths.

tests/fixtures.py
```diff
@@ -83,8 +83,10 @@ def setUp(self):
 
 
 # Except for python/mypy#731, prefer to define
-# FilesDef = Dict[str, Union['FilesDef', str]]
-FilesDef = Dict[str, Union[Dict[str, Union[Dict[str, str], str]], str]]
+# FilesDef = Dict[str, Union['FilesDef', str, bytes]]
+FilesDef = Dict[
+    str, Union[Dict[str, Union[Dict[str, Union[str, bytes]], str, bytes]], str, bytes]
+]
```
jaraco, line 89: Let's not complicate this type with bytes. I don't see anything below that can't be handled by strings.

tests/test_main.py
```diff
@@ -304,7 +312,12 @@ def test_packages_distributions_example2(self):
 
 
 class PackagesDistributionsTest(
-    fixtures.OnSysPath, fixtures.SiteDir, unittest.TestCase
+    fixtures.EggInfoPkg,
+    fixtures.EggInfoPkgPipInstalledNoModules,
+    fixtures.EggInfoPkgSourcesFallback,
```
jaraco, line 317: These fixtures aren't needed by other tests in this class. Perhaps it would be better to have a separate class or mechanism to inject these behaviors for the tests that rely on them.

tests/fixtures.py
```diff
@@ -211,6 +213,62 @@ def setUp(self):
         build_files(EggInfoPkg.files, prefix=self.site_dir)
 
 
+class EggInfoPkgPipInstalledNoModules(OnSysPath, SiteDir):
```
jaraco, line 216: Why is `NoModules` tested but not `WithModules` (presumably the more common use-case)? I'm particularly concerned that the "relative to egg-info directory" logic may prove problemmatic for package modules, especially since the egg-info directory isn't always in a reliable location relative to sources.

importlib_metadata/__init__.py
```diff
@@ -486,10 +495,35 @@ def _read_files_distinfo(self):
         text = self.read_text('RECORD')
         return text and text.splitlines()
 
-    def _read_files_egginfo(self):
+    def _read_files_egginfo_installed(self):
+        """
+        installed-files.txt might contain literal commas, so wrap
+        each line in quotes. Also, the entries in installed-files.txt
```
jaraco, line 501: This comment about 'literal commas' was okay when the implementation was small (barely different than _read_files_distinfo), but now that this interface is expanding to new use-cases, the description should probably give more precision around "returning CSV lines from installed-files.txt similar to lines from RECORD".

importlib_metadata/__init__.py
```diff
@@ -475,9 +475,18 @@ def make_file(name, hash=None, size_str=None):
 
         @pass_none
         def make_files(lines):
-            return list(starmap(make_file, csv.reader(lines)))
-
-        return make_files(self._read_files_distinfo() or self._read_files_egginfo())
+            return list(
+                filter(
+                    lambda package_path: package_path.locate().exists(),
+                    list(starmap(make_file, csv.reader(lines))),
```
jaraco, line 481: nit: This `list()` is no longer necessary.

importlib_metadata/__init__.py
```diff
@@ -475,9 +475,18 @@ def make_file(name, hash=None, size_str=None):
 
         @pass_none
         def make_files(lines):
-            return list(starmap(make_file, csv.reader(lines)))
-
-        return make_files(self._read_files_distinfo() or self._read_files_egginfo())
+            return list(
+                filter(
+                    lambda package_path: package_path.locate().exists(),
```
jaraco, line 480: Given the complexity, I wonder if it makes sense to separate this 'extant files' behavior from the rest.

tests/test_main.py
```diff
@@ -170,11 +170,19 @@ def test_metadata_loads_egg_info(self):
...
+    fixtures.DistInfoPkg,
+    unittest.TestCase,
+):
     def test_package_discovery(self):
         dists = list(distributions())
         assert all(isinstance(dist, Distribution) for dist in dists)
         assert any(dist.metadata['Name'] == 'egginfo-pkg' for dist in dists)
+        assert any(dist.metadata['Name'] == 'empty_egg-pkg' for dist in dists)
+        assert any(dist.metadata['Name'] == 'starved_egg-pkg' for dist in dists)
```
jaraco, line 185: Looking at the diff, I find it a little difficult to relate "PipInstalledNoModules" to "empty_egg" and "SourcesFallback" to "starved_egg". Is there a naming that would reflect the relationship more directly?

tests/test_main.py
```diff
@@ -170,11 +170,19 @@ def test_metadata_loads_egg_info(self):
         assert meta['Description'] == 'pôrˈtend'
 
 
-class DiscoveryTests(fixtures.EggInfoPkg, fixtures.DistInfoPkg, unittest.TestCase):
+class DiscoveryTests(
+    fixtures.EggInfoPkg,
+    fixtures.EggInfoPkgPipInstalledNoModules,
+    fixtures.EggInfoPkgSourcesFallback,
```
jaraco, line 176: I worry a bit about the complexity that this adds. I'm not sure what to do about it at this moment, but my initial instict is that these concerns are too entangled (suggesting that EggInfo and DistInfo were too entangled before, but since it was just two, it was acceptable). Maybe this approach is suitable, but I'll leave this flag for now to consider.

jaraco (commented):
Thanks @jherland for the contrib. I can see you took great care to be minimally invasive while addressing the core concerns. Well done, and kudos for your approach, especially considering the wonky structure of the fixtures and test parameterization. 

Overall, this approach looks acceptable. My main concern is with the non-metadata files as presented by installed-files. I have some style/quality concerns, but nothing serious. Feel free to reject or ignore any of my comments and I can address them to my satisfaction.

I'll want to spend a bit more time with this change before merging it in, but feel free to address some of my comments as your time permits.

importlib_metadata/__init__.py
jherland: Yes, that is one solution. There is already `.locate_file()` which is not far off, except that its implementation in `PathDistribution` precisely _removes_ the metadata directory that we need in this case.

More generally, I wonder if there is just a tad too much coupling between the `Distribution` and `PathDistribution` classes?

There is a mix of abstract and concrete methods in the `Distribution` base class, and several of the concrete methods seem fairly tightly bound to the `PathDistribution` subclass. For example, the `at()` static method creates and returns a `PathDistribution` instance outright, and the `metadata()`, `entry_points()`, `files()`, and `requires()` methods are all implemented with assumptions as to which files should be available and that the organization should resemble a dist-info or egg-info distribution.

Now, there is (AFAICS) only the one (`PathDistribution`) subclass of `Distribution` available in this project, and I don't really have enough experience with this project or its users to really imagine what _other_ `Distribution` subclasses would or could exist in the wild.

Still, I would suspect that a subclass that was sufficiently different from `PathDistribution` to rather prefer subclassing `Distribution` directly, would then find itself not merely implementing the `read_text()` an `locate_file()` abstract methods, but would probably also need to _re_implement several of `metadata()`, `entry_points()`, `files()`, or `requires()` as well.

Hence, I would raise the question whether some of these methods would be better off with their concrete implementations moved into the `PathDistribution` subclass? (Of course leaving abstract methods behind in `Distribution` where that makes sense.) This would grant these concrete methods direct access to the stuff they need inside `PathDistribution`, instead of having to add more interfaces to `Distribution` for stuff that really only makes sense for `PathDistribution`.

Still, this is a much bigger refactoring than I set out to do, and I would not feel comfortable starting this without consulting you.

tests/fixtures.py
jherland: I use the bytes version in `EggInfoPkgPipInstalledNoModules` to add a `top_level.txt` file that consists of a single newline:
```python
            # top_level.txt correctly reflects that no modules are installed
            "top_level.txt": b"\n",
```
Passing as a string here causes the newline to be stripped by the `DALS()` helper that is applied to string values in `build_files()`. I do test that the newline is preserved in `APITests.test_read_text()` (which is how I discovered the involvement of `DALS()`).

One thing I mentioned in #115 (but did not really touch in this PR) is the existence of a top_level.txt that is empty. This is currently treated as if the `top_level.txt` did not exist (i.e. we fall back to `_top_level_inferred()`). The `top_level.txt` with the single newline does reflect what `qiskit` ships with, and if we were to change the behavior in this area, then `EggInfoPkgPipInstalledNoModules` could serve as a relevant test case.

tests/test_main.py
jherland: Ack. I have to admit I'm not too familiar with how these fixtures interact with `unittest.TestCase`-style tests, and how I would proceed to apply the fixtures to one test method only.

Will instead move the method to a new `PackagesDistributionsEggTest` class.

tests/fixtures.py
jherland: Good catch, I'm adding a `EggInfoPkgPipInstalledNoToplevel` class now that does include a module that should be inferred via `installed-files.txt`. And there are indeed bugs/complications here. Will fix.

tests/test_main.py
jherland: Will rename them to "egg_with_module" and "sources_fallback"

importlib_metadata/__init__.py
jherland: Will remove.

importlib_metadata/__init__.py
jherland: Do you mean a simple refactoring like this?
```python
        @pass_none
        def make_files(lines):
            return list(starmap(make_file, csv.reader(lines)))

        @pass_none
        def skip_missing_files(package_paths):
            return list(filter(lambda path: path.locate().exists(), package_paths))

        return skip_missing_files(
            make_files(
                self._read_files_distinfo()
                or self._read_files_egginfo_installed()
                or self._read_files_egginfo_sources()
            )
        )
```

tests/test_main.py
jherland: Maybe it would look better to restructure the testdata (i.e. the various `DistInfo*` and `EggInfo*` classes in the `fixtures` module) using `@pytest.mark.parametrize(...)` instead of passing them as base classes to these tests? A major difference would then be that the tests are run against each of the package types in turn, instead of running them _once_ testing all packages types in one go. But there might be good reasons for why the tests are structured the current way?

tests/test_main.py
jherland: Done in [b391f77](https://github.com/python/importlib_metadata/pull/437/commits/b391f77d078b461da972226dbf6f0649e78eb5db)

tests/fixtures.py
jherland: Added `EggInfoPkgPipInstalledNoToplevel` in https://github.com/python/importlib_metadata/pull/437/commits/110f00d24304fdf446a640513f6bf7aef18ba9c3, and fixed associated issue in https://github.com/python/importlib_metadata/pull/437/commits/eeb2ed1593e25e4be3c7da8db073260a4ae8ec1e

importlib_metadata/__init__.py
jherland: Fixed in https://github.com/python/importlib_metadata/pull/437/commits/33eb7b4a0312017048d849cf1aef0321282d329d

importlib_metadata/__init__.py
jherland: Removed in https://github.com/python/importlib_metadata/pull/437/commits/61eca31a7456ccc39e3fda8e098497a4f28482b4

importlib_metadata/__init__.py
jherland: Did this in https://github.com/python/importlib_metadata/pull/437/commits/9b165a91af49bb11b7cf06b5661801f603aaeba6, can revisit if needed.

tests/test_main.py
jherland: Renamed in https://github.com/python/importlib_metadata/pull/437/commits/f62bf95d921fd8aca553813533f110d9010c2412 and https://github.com/python/importlib_metadata/pull/437/commits/a2dc88a299d6cb83052d91928b12cde0a6e5a2de

tests/test_main.py
jaraco: That's right. Although the test suite here leverages pytest, these same tests are ported into CPython where pytest is unavailable, so only unittest functionality is available. That's the motivation for these aberrant patterns. Elsewhere, I've started re-implementing `parameterization` similar to what pytest provides. I'm going to kick this can down the road for now and leave a refactoring for a later consideration.

tests/fixtures.py
jaraco: Oh! Thanks for the explanation. That's rather annoying that we have to hack in bytes support to bypass the indentation helper DALS. I think I'd rather address this concern head-on in order to retain simplicity of the FilesDef. I'm thinking something like a literal wrapper, `"top_level.txt": Literal("\n"),`, that DALS would ignore, or perhaps bypassing the file builder for this one file.

Or maybe this project should adopt [the technique](https://github.com/python/importlib_resources/blob/main/importlib_resources/tests/_path.py) I've used across other projects and which has been vendored into importlib_resources, but requires explicit invocation of DALS for every file. That version has bytes support and somehow doesn't require the messy type definition (maybe because of `# type: ignore`).

Given this mess, and given that your approach is converging on what FilesSpec does, let's go with this as implemented, and I'll revisit any lingering concerns I have separately.

importlib_metadata/__init__.py
jaraco: > Now, there is (AFAICS) only the one (`PathDistribution`) subclass of `Distribution` available in this project, and I don't really have enough experience with this project or its users to really imagine what _other_ `Distribution` subclasses would or could exist in the wild.

This project does aim to support an abstract concept of a `Distribution`, which could be a package served from a database or binary bundle or dynamically from memory. One downstream derived type I've [encountered](https://github.com/brechtm/rinohtype/pull/270) is [DynamicRinohDistribution](https://github.com/jaraco/rinohtype/blob/e7f1c2c6066303d86cff4105be0829f512059be2/src/rinoh/resource.py#L132).

I think you're right that a review/refactor may be in order, but is probably out of scope for this effort. I'll re-use your analysis in a new issue to track follow-up work.

tests/test_main.py
```diff
@@ -328,28 +339,76 @@ def test_packages_distributions_all_module_types(self):
...
-                        f'{i}-in-package/__init__.py,,\n'
-                        f'{i}-in-package/mod{suffix},,\n'
-                        for i, suffix in enumerate(suffixes)
-                    ),
+                    'RECORD': ''.join(f'{fname},,\n' for fname in filenames),
                 },
             },
             prefix=self.site_dir,
         )
+        fixtures.build_files({fname: "" for fname in filenames}, prefix=self.site_dir)
```
jaraco, line 365: I've noticed these changes conflict with other changes to main. Also, they seem somewhat independent of the scope of this issue (though I understand they may need to adapt). My plan is this:

- Back out the two commits that affect this test, alleviating the conflicts.
- In main, refactor the behavior to (a) ensure files are created and (b) derive the 'RECORD' from the FilesDef.
- Probably at the same time, harmonize the behavior of file creation with that of `importlib_resources` by relying on `jaraco.path.build`.

I've started on the work to support these behaviors in https://github.com/jaraco/jaraco.path/commit/13732c82dbe0e86cd13fb9cd656bc4509a37c610.

jaraco: Done in #446.

importlib_metadata/__init__.py
```diff
@@ -486,10 +497,43 @@ def _read_files_distinfo(self):
...
+        # which is not easily accessible from this base class...
+        subdir = getattr(self, '_path', None)
+        if not text or not subdir:
+            return
+        with contextlib.suppress(Exception):
+            ret = [
+                str((subdir / line).resolve().relative_to(self.locate_file('')))
+                for line in text.splitlines()
+            ]
+            return map('"{}"'.format, ret)
```
jaraco, line 524: I suspect this line ends up producing Windows path separators whereas the other `_read_files` methods produce `posixpath.sep`-arated paths (https://github.com/python/cpython/issues/103661).

jaraco: Confirmed. Solution in progress in https://github.com/python/importlib_metadata/pull/454.

PR #439, opened 2023-03-15
Update main.yml permissions

### LINKED ISSUE #438: Set github workflow to minimal permissions
I would like to suggest setting the permissions to the github workflows (currently the main.yml file) as read only on the [top level](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) and any write permission be given at the [run level](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idpermissions).

This is necessary due to a behavior of github workflow to grant to GITHUB_TOKEN write permissions to all types of permissions, regardless of they being used or not. In case of the workflow getting compromised, an attacker can exploit this permissions.

This can be seen in the Action run step "Set up job" such as https://github.com/python/importlib_metadata/actions/runs/4380969737/jobs/7668590805

image

Thus, it is both a recommendation from [OpenSSF Scorecard](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and the [Github](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions) to always use credentials that are minimally scoped.

Let me know if a PR is welcome and I'll submit it as soon as possible. 

Thanks!

Disclosure: I'm from Google, working with the OpenSSF to help many open source projects to increase their supply-chain security.

joycebrum:
### Changes

Closes #438 

Looking at tox documentation, it does not seem to need id-token. Also I've looked into https://github.com/jaraco/jaraco.develop/blob/main/jaraco/develop/create-github-release.py and the permissions seems to be only `metadata: read` (which is always read) and `contents: write`(granted to the job).

The other jobs seems to need only contents read, but I wasn't able to check due to test failings.

See what you think and if I may be missing something.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2,6 +2,9 @@ name: tests
 
 on: [push, pull_request]
 
+permissions:
+  contents: read
+
 env:
   # Environment variables to support color support (jaraco/skeleton#66):
   # Request colored output from CLI tools supporting it. Different tools
@@ -133,6 +136,8 @@ jobs:
     needs:
     - check
     if: github.event_name == 'push' && contains(github.ref, 'refs/tags/')
+    permissions:
+      contents: write
     runs-on: ubuntu-latest
 
     steps:


jaraco:
Superseded by jaraco/skeleton#76.

joycebrum:
Just a comment: although the used workflow have its permissions set to minimal scope, since is this workflow that creates the GITHUB_TOKEN, to all other commands/workflows it will have the write-all permission if no permission is set on the yml file.

I believe this PR would still be a add to supply-chain security, if possible, please reconsider.

Thanks!

FFY00:
@joycebrum the change was merged into https://github.com/jaraco/skeleton, and this repo was updated to the latest skeleton version, so your change already got pulled :sweat_smile: 

https://github.com/python/importlib_metadata/commits/main
https://github.com/python/importlib_metadata/commit/109f8c09ddb4904dc3f83307473520b2250ccb30

joycebrum:
Aaaah my bad. I didn't noticed it worked like that 😅. Thanks for the explanation.

PR #443, opened 2023-03-25
Filter excess entries

### LINKED ISSUE #442: metadata directories appearing in packages_distributions
I wonder if this adds the metadata directories into the output of `packages_distributions()` as import names. After this was merged, I'm seeing entries like these returned from `packages_distributions()`:
```
 'black-23.1.0.dist-info': ['black'],
 'filelock-3.10.0.dist-info': ['filelock'],
 'iniconfig-2.0.0.dist-info': ['iniconfig'],
 'more_itertools-9.1.0.dist-info': ['more-itertools'],
 'typing_extensions-4.5.0.dist-info': ['typing_extensions'],
```
IMHO, these should never occur here, as they are not importable modules.

_Originally posted by @jherland in https://github.com/python/importlib_metadata/issues/432#issuecomment-1475088318_

jaraco:
- Mark test as xfail as it's about to fail. Ref #442.
- Add test capturing missed expectation. Ref #442.
- 👹 Feed the hobgoblins (delint).
- Prefer all when asserting all.
- Filter non-identifiers from module names. Fixes #442.

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -360,5 +360,4 @@ class PackagesDistributionsTest(
         # All keys return from packages_distributions() should be valid
         # import names, which means that they must _at least_ be valid
         # identifiers:
-        for import_name in distributions.keys():
-            assert import_name.isidentifier(), import_name
+        assert all(import_name.isidentifier() for import_name in distributions.keys())
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -357,7 +357,8 @@ class PackagesDistributionsTest(
             assert distributions[f'in_namespace_{i}'] == ['all_distributions']
             assert distributions[f'in_package_{i}'] == ['all_distributions']
 
-        # All keys return from packages_distributions() should be valid import
-        # names, which means that they must _at least_ be valid identifiers:
+        # All keys return from packages_distributions() should be valid
+        # import names, which means that they must _at least_ be valid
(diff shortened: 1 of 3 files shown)


tests/test_main.py
```diff
@@ -350,6 +350,11 @@ def test_packages_distributions_all_module_types(self):
...
-            assert distributions[f'{i}-in-namespace'] == ['all_distributions']
-            assert distributions[f'{i}-in-package'] == ['all_distributions']
+            assert distributions[f'top_level_{i}'] == ['all_distributions']
+            assert distributions[f'in_namespace_{i}'] == ['all_distributions']
+            assert distributions[f'in_package_{i}'] == ['all_distributions']
+
+        # All keys return from packages_distributions() should be valid
+        # import names, which means that they must _at least_ be valid
+        # identifiers:
+        assert all(import_name.isidentifier() for import_name in distributions.keys())
```
jaraco, line 360: This assertion is incorrect. Module names are allowed not to be identifiers when invoked by runpy. For example, `py -m pip-run` is a valid invocation of a file called `pip-run.py`. We'll need to find a way to more precisely describe the expected inclusions/exclusions.

jaraco: Addressed in 7968a08.

jaraco:
Releasing as v6.2.1.

Issue #371, opened 2022-03-08
PackageMetadata.__getitem__ returns None


saaketp:
* The first problem is almost certainly a bug, the annotation for `__getitem__` says it returns a `str` but it returns `None` if the metadata doesn't exist. https://github.com/python/importlib_metadata/blob/v4.11.2/importlib_metadata/_meta.py#L15

  For example:
  ```python
  from importlib.metadata import metadata
  print(metadata("pip")["asdfg"] is None)  # True
  ```

(second problem moved to #374).

jaraco:
Hi @saaketp. Thanks for the report. I do believe this issue is a bug. I'm yet unsure if the proper fix is to update the protocol to match the implementation or to update the implementation to match the protocol. Historically, the metadata handling relied primarily on the interface of [email.message.EmailMessage](https://docs.python.org/3/library/email.message.html#email.message.EmailMessage). That documentation indicates that `None` is returned for missing values, so maybe the protocol should be updated to match.

jaraco:
Honestly, I'm conflicted. My instinct is that the current protocol definition is right and that `__getitem__` should raise a `KeyError` if a key is missing. I'm unsure why `EmailMessage` deviated from this typical `dict` behavior. @warsaw, do you know why `EmailMessage` does not raise a `KeyError` for missing headers? Do you have reason to think that `importlib.metadata` should continue that tradition or should align instead with dict and other typical behavior for `__getitem__`?

jaraco:
Separately, I'm disappointed that `mypy` was no help here. As much grief as mypy gives, nitpicking and requiring lots of redundant type hints, it's simultaneously failed here to identify the reported issue, despite the fact that [the return type is declared](https://github.com/python/importlib_metadata/blob/14cce75299645467adcd17352cb07caada32c444/importlib_metadata/__init__.py#L585) but the returned object is a [subclass of a class that violates that declaration](https://github.com/python/importlib_metadata/blob/14cce75299645467adcd17352cb07caada32c444/importlib_metadata/_adapters.py#L8). It's conceivable I've missed an important detail that would have allowed mypy to do this validation here.

Perhaps the issue is that [typeshed underdefines the return type](https://github.com/python/typeshed/blob/a5ddf0bf0d84beb9081fc3d3cfde2991585f1da5/stdlib/email/message.pyi#L34) as Any. Perhaps someone would like to fix that.

saaketp:
[...]

Issue #384, opened 2022-06-01
Should PackageMetadata.get be supported?


duckinator:
This isn't a feature request. I'm looking for a quick clarification, so other folks who encounter python/typeshed#7767 have a definitive answer.

---

In the wild, I've seen things like this:

```
from importlib.metadata import metadata
print(metadata("bork").get("Version"))
```

Both the [stdlib documentation](https://docs.python.org/3/library/importlib.metadata.html) and the [readthedocs documentation](https://importlib-metadata.readthedocs.io/en/latest/) never mention `.get()`, and `PackageMetadata` does not define it. Due to that, using it and running mypy returns `error: "PackageMetadata" has no attribute "get"`. I suspect I made a bad assumption when I used it, but if so it's one I've seen other folks make &mdash; so it seems worthwhile to get an explicit answer on whether leaving it out is deliberate.

**So, my question is this:**

Is `.get()` an implementation detail that folks latched on to and should migrate away from? Or should it be added to `PackageMetadata`?

rra:
The specific use case I have for using `.get()` is that sometimes I'm retrieving package metadata that may or may not exist (in support libraries that use package metadata for various purposes, for example). Without `.get()`, this would require a `try/catch` block around a hash dereference. That can be awkward, particularly when retrieving multiple bits of metadata at the same time and leaving the ones that aren't set as `None`.

jaraco:
tl;dr - yes, it's an implementation detail.

This is another case (similar to #371), where the duck-typing approach has left ambiguity. This project has provided a viable implementation but not a rigorously specified one. In more recent releases, `PackageMetadata` was introduced to be explicit about the supported interface. That answers the question in the title. It's an implementation detail that was not expected to be used.

That said, perhaps PackageMetadata should supply `.get()`. For the standard providers, because that method already exists in this implementation. The challenge is that if there are other providers that don't provide .get and the protocol is changed, that implementation won't be in compliance with the protocol. So we'll want a compelling case why `.get` is valuable.

What about using the `.json` accessor? It specifies a Dict object, which does have `.get`. Could that be suitable for your needs?

Or alternately, does `PackageMetadata` provide enough of `Mapping` to be able to be cast to a `Dict`? It seems it does, as it supplies all of the [abstract methods](https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes) of `Mapping`. That may mean that there's a straightforward way to create an adapter that would provide .get over a PackageMetadata.

[...]

PR #444, opened 2023-04-07
add .get() to the PackageMetadata protocol

dimbleby:
`__getitem__` warns that it will raise `KeyError` rather than returning `None` (I see #371 etc)

that's all very well but there's currently no comfortable way for users who anticipate missing values to write their code in a way that doesn't trigger that warning.  We seem to be steered towards
```python
import warnings
try:
    with warnings.catch_warnings():
        warnings.filterwarnings("ignore", "Implicit None .* KeyError", DeprecationWarning)
        value = metadata[key]
except KeyError:
    value = None
```
which sure seems like a lot of ceremony.

The natural way to do it would be just `value = metadata.get(key)`.  That method exists, but is not exposed by the protocol so if we try to use it then mypy shouts at us

Therefore add `get()` to that protocol (and testcases showing that it works).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,10 @@
+v6.2.0
+======
+
+* #384: ``PackageMetadata`` now stipulates an additional ``get``
+  method allowing for easy querying of metadata keys that may not
+  be present.
+
 v6.1.0
 ======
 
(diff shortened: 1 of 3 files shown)


dimbleby:
Just spotted previous discussion in #384.  Conversation there doesn't sound too opposed to this.

IMO it's a very natural thing to have on the API, perhaps the new annoyance with the hard-to-avoid warning, alongside an actual merge request, is enough to settle the matter...

jaraco (approved):
Looks good. I'll just add a changelog entry and we can roll this out.

Issue #452, opened 2023-04-19
Cannot infer top-level imports from packages installed by PDM w/`install.cache=true` (aka. symlinked packages)


jherland:
[PDM](https://pdm.fming.dev/) has an alternative to install packages (enabled with `pdm config install.cache true`) where the package is first installed into a local cache (typically under `$HOME/.cache/pdm/packages/`), and then the package is installed into the target virtualenv, but with symlinks pointing back to the package in the cache. Here is an example:

```
$ pdm config install.cache true
$ pdm init
[...]
$ pdm add colorama
[...]
$ ls -al .venv/lib/python3.10/site-packages/colorama*
lrwxrwxrwx 1 jherland users   79 Apr 19 03:03 .venv/lib/python3.10/site-packages/colorama -> /home/jherland/.cache/pdm/packages/colorama-0.4.6-py2.py3-none-any/lib/colorama

.venv/lib/python3.10/site-packages/colorama-0.4.6.dist-info:
total 44
drwxr-xr-x 3 jherland users  4096 Apr 19 03:03 .
drwxr-xr-x 4 jherland users  4096 Apr 19 03:03 ..
drwxr-xr-x 2 jherland users  4096 Apr 19 03:03 licenses
-rw-r--r-- 1 jherland users 17158 Apr 19 03:03 METADATA
-rw-r--r-- 1 jherland users   412 Apr 19 03:03 RECORD
-rw-r--r-- 1 jherland users    66 Apr 19 03:03 REFER_TO
-rw-r--r-- 1 jherland users   105 Apr 19 03:03 WHEEL
$ cat .venv/lib/python3.10/site-packages/colorama-0.4.6.dist-info/RECORD
colorama-0.4.6.dist-info/METADATA,sha256=e67SnrUMOym9sz_4TjF3vxvAV4T3aF7NyqRHHH3YEMw,17158
colorama-0.4.6.dist-info/WHEEL,sha256=cdcF4Fbd0FPtw2EMIOwH-3rSOTUdTCeOSXRMD1iLUb8,105
colorama-0.4.6.dist-info/licenses/LICENSE.txt,sha256=ysNcAmhuXQSlpxQL-zs25zrtSWZW6JEQLkKIhteTAxg,1491
colorama-0.4.6.dist-info/REFER_TO,sha256=b2lWg3Osknh6HODITiPmZW55TlJ8okVt-8k6rf3kdpc,66
colorama-0.4.6.dist-info/RECORD,,
colorama,,
[...]

PR #453, opened 2023-04-19
Detecting and handling PDM's symlinked packages

jherland:
This is accompanying issue #452: The first commit adds a test case to demonstrate the issue, and the second attempt is a first, rough, attempt at fixing it.

I am unsure if a proper fix should only have `_top_level_inferred()` recurse/walk into symlinked directories, or if the recurse/walk should also apply to `Distribution.files()`.

Commits:
- Add test to demonstrate issue with symlinked packages
- Attempt to fix issue with symlinked packages


jaraco:
> I am unsure if a proper fix should only have `_top_level_inferred()` recurse/walk into symlinked directories, or if the recurse/walk should also apply to `Distribution.files()`.

My instinct is that `Distribution.files()` should be naive and return the files for the virtual package as installed and not resolve to the linked content - at least until such a time that linked packages are more formally defined and accepted as a standard in the Python ecosystem.

PR #454, opened 2023-04-22
Fix path handling and masked errors on Windows for installed-files.txt

jaraco:
- Resolve the located directory and remove suppression of Exceptions. Ref python/cpython#103661.
- Wrap 'subdir/line' in PosixPath to ensure the output uses posix path separators. Ref python/cpython#103661.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,10 @@
+v6.5.1
+======
+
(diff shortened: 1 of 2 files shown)


jaraco:
Note that after the first commit, the error uncovered by Python buildbots in python/cpython#103661 is revealed.

PR #460, opened 2023-06-21
add typing for simple functions

Gryfenfer97:
I just added typing to simple functions in `__init__.py` based on [the stub file of importlib.metadata](https://github.com/python/typeshed/blob/main/stdlib/importlib/metadata/__init__.pyi)


jaraco:
Looks great. Thanks for the contrib!

PR #465, opened 2023-07-30
Provide `Distribution.origin` reflecting content of direct_url.json.

jaraco:
Partial implementation of #404.

### DIFF
--- a/importlib_metadata/__init__.py
+++ b/importlib_metadata/__init__.py
@@ -3,8 +3,10 @@ import re
 import abc
 import csv
 import sys
+import json
 import zipp
 import email
+import types
 import inspect
 import pathlib
 import operator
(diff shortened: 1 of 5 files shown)


makslevental:
Any blockers on this? Would be nice to have.

jaraco:
diffcov is reporting this failure:

```
importlib_metadata/__init__.py (71.4%): Missing lines 632,635
```

jaraco:
I don't understand how the lines aren't covered, given the test specifically exercises the `.origin` property.

jaraco:
Oh, maybe the issue is the test isn't derived from `unittest.TestCase` so isn't discovered.

jaraco:
> Oh, maybe the issue is the test isn't derived from `unittest.TestCase` so isn't discovered.

Yep. That was it... and demonstrates that the test is not in fact capturing the expectation. Hooray for diffcov for catching an otherwise missed expectation.

makslevental:
funny i don't remember what i wanted this for but thanks!

PR #467, opened 2023-08-31
Upgrade GitHub Actions

cclauss:
https://github.com/actions/checkout/releases
https://github.com/actions/setup-python/releases

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -108,11 +108,11 @@ jobs:
   diffcov:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         with:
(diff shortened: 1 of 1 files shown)


jaraco:
Thanks for catching this concern. This project builds on jaraco/skeleton, which already has the bumps. I want to investigate when and why this project diverged to see if it was for some good reason.

jaraco:
Oh! It's the `diffcov` job, which doesn't exist upstream. Probably this project should either drop diffcov or maybe move it upstream, perhaps integrated into the test suite (instead of as a separate job). Otherwise, this job (and the same one in `importlib_resources`) will be a lingering maintenance concern.

jaraco:
Superseded by b02bf32bae729d53bdb7c9649d6ec36afdb793ee.

Issue #468, opened 2023-09-08
Invalid annotation in `SimplePath` protocol


sobolevn:
Hi! 

I was working on `typeshed`'s types and we (well, @AlexWaygood) found that annotation of `__truediv__` is not valid: https://github.com/python/importlib_metadata/blob/8e7b9689fc38ef3e834265fd78423a9933a33607/importlib_metadata/_meta.py#L55-L56

Here's a small repro: https://mypy-play.net/?mypy=latest&python=3.11&gist=2f16bbbbaf5a1330cbbcae76b5987868

```python
from typing import TypeVar

T = TypeVar('T')

class A:
    def method(self, other: T | str) -> T: ...

a: A
reveal_type(a.method(1))  # N: Revealed type is "builtins.int"
reveal_type(a.method('a'))  # N: Revealed type is "<nothing>"
```

I propose using an `@overload` instead:

```python

from typing import TypeVar, overload

T = TypeVar('T')

class A:
    @overload
    def method(self, other: str) -> str: ...
    @overload
    def method(self, other: T) -> T: ...

a: A
reveal_type(a.method(1))  # N: Revealed type is "builtins.int"
reveal_type(a.method('a'))  # N: Revealed type is "builtins.str"
```

Playground: https://mypy-play.net/?mypy=latest&python=3.11

I will send a PR with the fix :)

sobolevn:
Or we can just drop `str` part :)

jaraco:
Based on a comment in #469, it sounds as if this concern isn't in fact a problem because `SimplePath` itself is parameterized by `_T`. For now, I'm marking this as invalid, but happy to revisit, especially if someone can demonstrate for me how the current definition manifests problematic behavior.

PR #469, opened 2023-09-08
Fix type annotation in `SimplePath.__truediv__`

sobolevn:
Closes https://github.com/python/importlib_metadata/issues/468

### DIFF
--- a/importlib_metadata/_meta.py
+++ b/importlib_metadata/_meta.py
@@ -52,7 +52,7 @@ class SimplePath(Protocol[_T]):
     def joinpath(self, other: Union[str, _T]) -> _T:
         ...  # pragma: no cover
 
-    def __truediv__(self, other: Union[str, _T]) -> _T:
+    def __truediv__(self, other: _T) -> _T:
         ...  # pragma: no cover
 
     @property


dimbleby:
shouldn't `joinpath` agree with `__truediv__`?

importlib_metadata/_meta.py
```diff
@@ -52,7 +52,7 @@ class SimplePath(Protocol[_T]):
     def joinpath(self, other: Union[str, _T]) -> _T:
         ...  # pragma: no cover
 
-    def __truediv__(self, other: Union[str, _T]) -> _T:
+    def __truediv__(self, other: _T) -> _T:
```
jaraco, line 55: I'm not sure this annotation is correct. A SimplePath could accept a `str` as input and return a SimplePath. I understand that `pathlib.Path` is parameterized (the subtype is either `str` or `bytes`), but maybe SimplePath should just assume `str`. The only thing I'm uncertain about is if existing uses might utilize a `Path[bytes]`. That is, maybe we should eliminate `T` and replace it with `SimplePath`... so `__truediv__` accepts `other: Union[str, 'SimplePath']`.

jaraco:
Based on the discussion, my understanding is that the status quo is probably acceptable and thus we'll choose to reject this change (and its unresolved concerns). Thanks anyhow for the proposal. If you wish to revisit the concern, let's be sure to answer the open concerns about correctness and consistency.

PR #482, opened 2024-01-05
Add support for egg packages with files outside site-packages

### LINKED ISSUE #455: Uncaught Exception when reading Egg information
Hello,

I manage a project using poetry and since importlib-metadata `6.5.1` my install has been failing with the following error. I'm trying to install an updated version of `ansible` while a current one is already installed. I think the issue is that pathlib is having trouble finding a relative path between the ansible binary `/usr/local/bin/ansible` and site-packages.
```
ansible [core 2.13.9]
  config file = None
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.8/site-packages/ansible
  ansible collection location = /root/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/local/bin/ansible
  python version = 3.8.16 (default, Feb 11 2023, 02:53:15) [GCC 10.2.1 20210110]
  jinja version = 3.1.2
  libyaml = True
---
00:40:59  #12 42.71   ValueError
00:40:59  #12 42.71 
00:40:59  #12 42.71   '/usr/local/bin/ansible' does not start with '/usr/local/lib/python3.8/site-packages'
00:40:59  #12 42.71 
00:40:59  #12 42.71   at /usr/lib64/python3.8/pathlib.py:908 in relative_to
00:40:59  #12 42.83        904│         n = len(to_abs_parts)
00:40:59  #12 42.83        905│         cf = self._flavour.casefold_parts
00:40:59  #12 42.83        906│         if (root or drv) if n == 0 else cf(abs_parts[:n]) != cf(to_abs_parts):
00:40:59  #12 42.83        907│             formatted = self._format_parsed_parts(to_drv, to_root, to_parts)
00:40:59  #12 42.83     →  908│             raise ValueError("{!r} does not start with {!r}"
00:40:59  #12 42.83        909│                              .format(str(self), str(formatted)))
00:40:59  #12 42.83        910│         return self._from_parsed_parts('', root if n == 1 else '',
00:40:59  #12 42.83        911│                                        abs_parts[n:])
00:40:59  #12 42.83        912│ 
```
This error used to be caught by a way to broad broad `contextlib.suppress(Exception)` suppression which @jaraco's recent [PR](https://github.com/python/importlib_metadata/pull/454) removed. I'm not that familiar with pythons package/importing mechanisms but could we add back a specific `suppress(ValueError)` to [_read_files_egginfo_installed](https://github.com/python/importlib_metadata/blob/be58651120fb65d0fadf93dcc5f609b300968704/importlib_metadata/__init__.py#L540)? We could also completely avoid the error by manually loop through and checking the path  before finding its relative. I'm happy to open a PR with either change

dan-blanchard:
This should fix #455. I slightly modified the patch suggested by @jaraco in https://github.com/python/importlib_metadata/issues/455#issuecomment-1837571288 so that we don't have to call `self.locate_file('').resolve()` for `name` in `text.splitlines()`, and fixed one type complaint I was getting locally.

I'm not sure if the test I added works properly (which is one of the reasons I made this PR), because I'm running into a weird `tox` issue locally.

### DIFF
--- a/importlib_metadata/__init__.py
+++ b/importlib_metadata/__init__.py
@@ -562,14 +562,30 @@ class Distribution(DeprecatedNonAbstract):
         if not text or not subdir:
             return
 
+        site_path = self.locate_file('').resolve()
         paths = (
-            (subdir / name)
-            .resolve()
-            .relative_to(self.locate_file('').resolve())
-            .as_posix()
+            self._relative_to(
+                (subdir / name).resolve(),
+                site_path,
+            ).as_posix()
             for name in text.splitlines()
         )
         return map('"{}"'.format, paths)
+
+    def _relative_to(self, path, root):
+        """
+        Workaround for https://bugs.python.org/issue23082 where ".."
+        isn't added by pathlib.Path.relative_to() when path is not
+        a subpath of root.
+
+        One example of such a package is dask-labextension, which uses
+        jupyter-packaging to install JupyterLab javascript files outside
+        of site-packages.
+        """
+        try:
+            return path.relative_to(root)
+        except ValueError:
+            return pathlib.Path(os.path.relpath(path, root))
 
     def _read_files_egginfo_sources(self):
         """
--- a/importlib_metadata/_meta.py
+++ b/importlib_metadata/_meta.py
@@ -70,3 +70,6 @@ class SimplePath(Protocol):
 
     def exists(self) -> bool:
         ...  # pragma: no cover
+
+    def resolve(self) -> bool:
+        ...  # pragma: no cover
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -252,6 +252,40 @@ class EggInfoPkgPipInstalledNoToplevel(OnSysPath, SiteBuilder):
     }
 
 
+class EggInfoPkgPipInstalledExternalDataFiles(OnSysPath, SiteBuilder):
+    files: FilesSpec = {
+        "egg_with_module_pkg.egg-info": {
+            "PKG-INFO": "Name: egg_with_module-pkg",
+            # SOURCES.txt is made from the source archive, and contains files
+            # (setup.py) that are not present after installation.
+            "SOURCES.txt": """
+                egg_with_module.py
+                setup.py
+                egg_with_module.json
+                egg_with_module_pkg.egg-info/PKG-INFO
+                egg_with_module_pkg.egg-info/SOURCES.txt
+                egg_with_module_pkg.egg-info/top_level.txt
+            """,
+            # installed-files.txt is written by pip, and is a strictly more
+            # accurate source than SOURCES.txt as to the installed contents of
+            # the package.
+            "installed-files.txt": """
+                ../../../etc/jupyter/jupyter_notebook_config.d/egg_with_module.json
+                ../egg_with_module.py
+                PKG-INFO
+                SOURCES.txt
+                top_level.txt
+            """,
+            # missing top_level.txt (to trigger fallback to installed-files.txt)
+        },
+        "egg_with_module.py": """
+            def main():
+                print("hello world")
+            """,
+    }
+
+
+
 class EggInfoPkgPipInstalledNoModules(OnSysPath, SiteBuilder):
     files: FilesSpec = {
         "egg_with_no_modules_pkg.egg-info": {
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -29,6 +29,7 @@ class APITests(
     fixtures.EggInfoPkg,
     fixtures.EggInfoPkgPipInstalledNoToplevel,
     fixtures.EggInfoPkgPipInstalledNoModules,
+    fixtures.EggInfoPkgPipInstalledExternalDataFiles,
     fixtures.EggInfoPkgSourcesFallback,
     fixtures.DistInfoPkg,
     fixtures.DistInfoPkgWithDot,


dan-blanchard:
The "weird tox issue" turns out to be an issue with pytest-perf, I think.

I keep getting errors like:

```
E           subprocess.CalledProcessError: Command '('/opt/homebrew/Caskroom/mambaforge/base/envs/py312importlib/bin/python3.12', '-m', 'pip', 'install', '-t', PosixPath('/var/folders/vz/khc_rx815vv9rhbclb1bl2hc0000gn/T/pip-run-x6yeau2d'), 'importlib_metadata.git[perf]@git+git@github.com:dan-blanchard/importlib_metadata.git')' returned non-zero exit status 1.

/opt/homebrew/Caskroom/mambaforge/base/envs/py312importlib/lib/python3.12/subprocess.py:413: CalledProcessError
---------------------------------------------------------------------------------- Captured stderr call -----------------------------------------------------------------------------------
ERROR: Invalid requirement: 'importlib_metadata.git[perf]@git+git@github.com:dan-blanchard/importlib_metadata.git'
Hint: It looks like a path. File 'importlib_metadata.git[perf]@git+git@github.com:dan-blanchard/importlib_metadata.git' does not exist.

```

when trying to run tests locally.

jaraco:
> Invalid requirement: 'importlib_metadata.git[perf]@git+git@github.com:dan-blanchard/importlib_metadata.git'

You may want to report this issue to [pytest-perf](/jaraco/pytest-perf). I notice that `importlib_metadata.git[perf]` should be `importlib_metadata[perf]`, and probably what's happening is the plugin is inferring the package name from the upstream URL, but since you're using the SSH form which may (or may not) expect the `.git` suffix.

That's one of the reasons I avoid the SSH protocol - because it tends to add noise like `.git` to URLs.

Trying to install it without the package name, I still get the same error:

```
 @ pip-run git+git@github.com:dan-blanchard/importlib_metadata.git
ERROR: Invalid requirement: 'git+git@github.com:dan-blanchard/importlib_metadata.git'
Hint: It looks like a path. File 'git+git@github.com:dan-blanchard/importlib_metadata.git' does not exist.
```

Perhaps it's the case that the syntax pip expects is different than what git supplies. Indeed, the [pip docs](https://pip.pypa.io/en/stable/topics/vcs-support/#supported-vcs) indicate that the URL should be `git+ssh://`, so probably pytest-perf needs to add support for translating a git remote to a pip URL for SSH remotes.

If you want to work around the issue, you can use an https remote (e.g. `git clone https://github.com/$(username)/importlib_metadata`).

Let me see if I can resolve the conflicts and get the tests running again.

importlib_metadata/__init__.py
```diff
@@ -562,15 +562,31 @@ def _read_files_egginfo_installed(self):
...
+                (subdir / name).resolve(),
+                site_path,
+            ).as_posix()
             for name in text.splitlines()
         )
         return map('"{}"'.format, paths)
 
+    def _relative_to(self, path, root):
+        """
+        Workaround for https://bugs.python.org/issue23082 where ".."
```
jaraco, line 577:
```suggestion
        Workaround for https://github.com/python/cpython/issues/67271 where ".."
```

jaraco:
Thanks for this. Looks like it's pretty close.

I'm a little concerned that this change introduces more reliance on the assumption that the package is on disk, when it could be in a zip file or elsewhere.

dan-blanchard:
I believe all the test failures were mypy related so I added a `cast()` to fix/silence it.

jaraco:
The issue with pytest-perf supporting SSH URLs was addressed in https://github.com/jaraco/pytest-perf/issues/15.

importlib_metadata/__init__.py
```diff
@@ -440,7 +440,7 @@ def at(path: str | os.PathLike[str]) -> Distribution:
         :param path: a string or path-like object
         :return: a concrete Distribution instance for the path
         """
-        return PathDistribution(pathlib.Path(path))
+        return PathDistribution(cast(SimplePath, pathlib.Path(path)))
```
jaraco, line 443: I notice that the commit message says this change is unrelated to the PR, but it does seem to be related, as tests are passing in main but not against this PR rebased to main. I'd like to understand better why this change is necessary (and that it's correct).

importlib_metadata/__init__.py
```diff
@@ -562,15 +562,31 @@ def _read_files_egginfo_installed(self):
...
+
+        One example of such a package is dask-labextension, which uses
+        jupyter-packaging to install JupyterLab javascript files outside
+        of site-packages.
+        """
+        try:
+            return path.relative_to(root)
+        except ValueError:
+            return pathlib.Path(os.path.relpath(path, root))
+
```
jaraco, line 589: It looks like the bug was fixed in Python 3.9. We should move this workaround to .compat/py38 so it can be removed when no longer needed and can be omitted when porting this code into CPython.

jaraco: Oh, so I tried extracting this function into a wrapper that's only applied on Python 3.9 and later, but then the ValueError is emitted. So now I no longer trust the comment, which indicates that this is a workaround for a bug that's closed.

jaraco: Aha. So I found that there is https://github.com/python/cpython/issues/84538, which captures the underlying issue, and the fix for which adds a `walk_up` parameter.

jaraco: Fixed in 1584b96. Now the diff for Python 3.12+ is simply to use `walk_up=True`.

importlib_metadata/__init__.py
jaraco: Reverting 4a3782d11895c22fd443c67175ab7352db2a769b, I see the following error from mypy:

```
443: error: Argument 1 to "PathDistribution" has incompatible type "Path"; expected "SimplePath"  [arg-type]
443: note: Following member(s) of "Path" have conflicts:
443: note:     Expected:
443: note:         def __truediv__(self, str | PathLike[str], /) -> SimplePath
443: note:     Got:
443: note:         def __truediv__(self, str | PathLike[str], /) -> Path
443: note:     Expected:
443: note:         def joinpath(self, other: str | PathLike[str]) -> SimplePath
443: note:     Got:
443: note:         def joinpath(self, *other: str | PathLike[str]) -> Path
443: note:     <2 more conflict(s) not shown>
```

It seems somehow `Path` isn't considered a `SimplePath`. Oh, perhaps it's the introduction of the `def resolve()` to the `SimplePath` protocol.

importlib_metadata/_meta.py
```diff
@@ -65,3 +65,5 @@ def read_text(self, encoding=None) -> str: ...  # pragma: no cover
     def read_bytes(self) -> bytes: ...  # pragma: no cover
 
     def exists(self) -> bool: ...  # pragma: no cover
+
+    def resolve(self) -> bool: ...  # pragma: no cover
```
jaraco, line 69: Why was this added?

jaraco: Not only does this change seem unnecessary, it's also incorrect. `.resolve` doesn't return a `bool`. It returns another `SimplePath`.

Moreover, I suspect other implementations of `SimplePath` don't implement resolve.

In particular, the [zipfile.Path](https://docs.python.org/3/library/zipfile.html#path-objects) is meant to be another implementation of `SimplePath` and is used by `PathDistribution` to support packages in a zip file. It also looks like `zipfile.Path` doesn't support `.relative_to` or `.as_posix`, so adding `.resolve()` to the protocol won't be sufficient.

It does appear that the `_read_files_egginfo_installed` routine has never worked for zipfile.Paths.

And if nobody's asking for it to do so, I'm not going to worry about it. Eggs are deprecated and support is best-effort.

jaraco: I've found that removing this change allows the tests to pass, so I'm going to assume this change is not needed and proceed without it. Removed in e4d1dcca72.

importlib_metadata/__init__.py
jaraco: Yes, that was the case. Removing the `def resolve` from the `SimplePath` spec resolves the issue.

jaraco:
Released as v7.2.1.

PR #483, opened 2024-01-30
Add typing for load and select

Danyal-Faheem:
**Changes**

- Added typing for the `Entrypoint.load()` and the `Entrypoints.select()` method.

As the `Entrypoint.load()` returns more than one type depending on the value of the `value` variable, it was necessary to set it as `Any`.

### DIFF
--- a/importlib_metadata/__init__.py
+++ b/importlib_metadata/__init__.py
@@ -33,7 +33,7 @@ from contextlib import suppress
 from importlib import import_module
 from importlib.abc import MetaPathFinder
 from itertools import starmap
-from typing import Iterable, List, Mapping, Optional, Set, cast, Any
+from typing import Any, Iterable, List, Mapping, Optional, Set, cast
 
 __all__ = [
     'Distribution',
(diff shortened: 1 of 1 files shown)


importlib_metadata/__init__.py
```diff
@@ -33,7 +33,7 @@
 from importlib import import_module
 from importlib.abc import MetaPathFinder
 from itertools import starmap
-from typing import Iterable, List, Mapping, Optional, Set, cast
+from typing import Iterable, List, Mapping, Optional, Set, cast, Any
```
jaraco, line 36:
```suggestion
from typing import Any, Iterable, List, Mapping, Optional, Set, cast
```

jaraco (approved):

jaraco (changes requested):
Looks like this change introduces a couple of new errors:

```
________________________ importlib_metadata/__init__.py ________________________
184: error: Item "None" of "Optional[Match[str]]" has no attribute "group"  [union-attr]
185: error: Item "None" of "Optional[Match[str]]" has no attribute "group"  [union-attr]
```

jaraco (approved):

Issue #486, opened 2024-05-02
API incompatibility with `importlib.metadata` (or at least the API is not type-safe?)


abravalheri:
When 3rd-party `MetaPathFinder` are implemented using `importlib.metadata`, the return types in `importlib_metadata` are not respected, which causes the API to behave in an unexpected way (with unexpected errors).

This is an example identified in https://github.com/pypa/pyproject-hooks/pull/195#issuecomment-2088695300 and https://github.com/pypa/setuptools/issues/4338:

```bash
mkdir -p /tmp/stash/private_path/_test-0.0.1.dist-info
cat <<EOF > /tmp/stash/private_path/_test-0.0.1.dist-info/METADATA
Name: _test
Version: 0.0.1
EOF

cat <<EOF > /tmp/stash/private_path/_test-0.0.1.dist-info/entry_points.txt
[_test.importlib_metadata]
hello = world
EOF

cat <<EOF > /tmp/stash/install_finder.py
import sys
from importlib.machinery import PathFinder
from importlib.metadata import DistributionFinder, MetadataPathFinder


class _ExampleFinder:
    def __init__(self, private_path):
        self.private_path = private_path


    def find_spec(self, fullname, _path, _target=None):
        if "." in fullname:
            # Rely on importlib to find nested modules based on parent's path
            return None

        # Ignore other items in _path or sys.path and use private_path instead:
        return PathFinder.find_spec(fullname, path=self.private_path)

    def find_distributions(self, context=None):
        context = DistributionFinder.Context(path=self.private_path)
        return MetadataPathFinder.find_distributions(context=context)


sys.meta_path.insert(0, _ExampleFinder(["/tmp/stash/private_path"]))
EOF

cd /tmp/stash/
python3.8 -m venv .venv
.venv/bin/python -m pip install -U importlib-metadata
.venv/bin/python
```

```python
>>> import install_finder
>>> from importlib_metadata import distribution
>>> distribution("_test").entry_points.select(group="_test.importlib_metadata")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'list' object has no attribute 'select'
```

The expected behaviour as per API documentation would be:
```python
>>> distribution("_test").entry_points.select(group="_test.importlib_metadata")
EntryPoints((EntryPoint(name='hello', value='world', group='_test.importlib_metadata'),))
```

It seems that the origin of this problem is a little "lie" in the API definition:

Instead of

[...]

PR #487, opened 2024-05-02
Adopt more realistic type annotations

abravalheri:
Motivated by #486.

This PR just provides more realistic type annotations (a bit more correct technically speaking) to improve the type checking soundness of the package.

The limitations discussed in #486 are still present, but at least type checkers would help to notify users when they are trying to use APIs that are not safe (not necessary all the time, but better than now).

It is a bit of a pain, but these annotation should be a bit more sound than they currently are.


jaraco:
Thanks for the proposal. I'm going to have to think about this more.

This library is merged back into CPython as `importlib.metadata` so these annotations won't make sense when that happens. Moreover, this is the first introduction of references to `importlib.metadata` in this library, so I'm unsure what the implications will be.

Is there any way to achieve the objectives either by:

- adjusting the patch to be universal (so it's meaningful in both importilb_metadata and importlib.metadata), or
- moving the backport-specific concerns to a `./compat/*` module that would be omitted downstream in CPython?

abravalheri:
Thanks @jaraco for the code review.

> Is there any way to achieve the objectives either by:
> 
> * adjusting the patch to be universal (so it's meaningful in both importilb_metadata and importlib.metadata), or

Unfortunately I don't think so.
`importlib_metadata` has a backwards compatibility problem that simply disappears when it becomes `importlib.metadata`.

In this sense only `importlib_metadata` need these contrived type annotations to inform end-users that they can get back legacy objects when using the API.


> * moving the backport-specific concerns to a `./compat/*` module that would be omitted downstream in CPython?

I gave it a try... don't know if that is exactly what you had in mind, but I tried to do my best.
Please note that the annotations are not specific to a version of Python, but always will be necessary while `importlib_metadata` exists in PyPI. In the future I believe they may be useful during development (e.g. by identifying places where legacy objects are retrieved and preventing incompatible methods to be called on them).

I am not good with names, so if you have suggestions regarding names please let me know.

importlib_metadata/compat/stdlib.py
```diff
@@ -0,0 +1,27 @@
...
+    from typing_extensions import TypeAlias
+
+    from .. import Distribution, PackagePath, _meta
+
+    if sys.version_info >= (3, 10):
+        from importlib.metadata import PackageMetadata as _legacy_Metadata
+    else:
+        from email.message import Message as _legacy_Metadata
+
+    _PackageMetadataOrLegacy: TypeAlias = Union[_legacy_Metadata, _meta.PackageMetadata]
```
abravalheri, line 23: I previously tried a more modern syntax, but apparently `from __future__ import annotations` does not work when defining aliases (i.e. you cannot have `|` or `list[...]` on the right side of an assignment).

https://github.com/python/importlib_metadata/actions/runs/9643043347/job/26592160998?pr=487

jaraco: Sometimes in this case, I would suggest splitting the definition on Python version, but given that anyone could try to implement the more modern syntax and immediately get feedback that it's invalid on Python 3.9, I'm perfectly fine with just leaving it as-is. It might mean that this syntax lingers longer than it has to after dropping support for Python 3.9, but that's an acceptable tradeoff for having a unified definition.

importlib_metadata/compat/stdlib.py
```diff
@@ -0,0 +1,27 @@
...
+from typing import TYPE_CHECKING, List, Union
+
+if TYPE_CHECKING:
+    # Avoid circular imports
+
+    from importlib import metadata as _legacy
+
+    from typing_extensions import TypeAlias
+
+    from .. import Distribution, PackagePath, _meta
```
jaraco, line 16: Shouldn't these come from `importlib.metadata`?

jaraco: Aah. Okay. Yes.

The term legacy was throwing me off. I was thinking surely `importlib_metadata` would be the legacy. Thanks for confirming.

jaraco:
As discussed in #486, I believe the direction we're going is to unify the the interfaces (through #505 plus new protocols).

Issue #495, opened 2024-07-19
Vulnerability issue CVE-2024-5569


karistom:
zipp vulnerability issue CVE-2024-5569 caused by `importlib-metadata`.
https://github.com/advisories/GHSA-jfmj-5v4g-7637
Patched zipp version is 3.19.1.

We are using Python 3.11.0, importlib-metadata is calling zipp-3.15.0.
We even tried to use the latest version (importlib-metadata>=8.0.0), but it is still calling old version of zipp.

karistom:
Please check this code.
https://github.com/python/importlib_metadata/blob/f3901686abc47853523f3b211873fc2b9e0c5ab5/pyproject.toml#L21

jaraco:
Thanks for the report.

My initial instinct is not to update the lower bound. I usually consider the lower bound for a dependency to indicate versions that are compatible, and since importlib_metadata is still compatible with zipp>=0.5, that seems like the best lower bound, even though some users may choose to use a newer version for performance or security or other reasons.

Moreover, the vulnerability that affected zipp was for specially-crafted zip files, which is an even narrower scope for importlib_metadata, where the crafted zip file would also have to be present on the Python path.

Furthermore, it's not feasible in general to find all projects dependent on zipp and ensure they're pinning to a particular version.

> We are using Python 3.11.0, importlib-metadata is calling zipp-3.15.0.

That may be the case in some scenarios, but the default behavior for a new system will be to install the newest version of zipp. It's only in old, stale environments where the older version will be cached.

----

Overall, this leads me to conclude that we should not alter the declared dependency.

Can you make the case to the contrary?

karistom:
Hi James, in our security report, `importlib-metadata` is keep calling old version of `zipp`. Unless we specify the latest version `3.19.1` in `importlib-metadata` it will be calling old version which is causing security issue.

jaraco:
> Hi Jason, in our security report, `importlib-metadata` is keep calling old version of `zipp`. Unless we specify the latest version `3.19.1` in `importlib-metadata` it will be calling old version which is causing security issue.

importlib-metadata will call whatever version of `zipp` is installed, putting the onus of updating `zipp` on the system integrator to keep dependencies up-to-date. For example, `pip install --upgrade --upgrade-strategy eager importlib_metadata` or simply `pip install --upgrade zipp` will update zipp to the latest release and address the security issue.

I don't believe the principle of updating the pin of a dependent package for any security fix in a dependency generalizes. 

[...]

PR #496, opened 2024-07-22
zipp version CVE-2024-5569

karistom:
To address [CVE-2024-5569](https://github.com/advisories/GHSA-jfmj-5v4g-7637) vulnerability issue,
updated zipp version.

Related git issue: https://github.com/python/importlib_metadata/issues/495

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,8 @@ classifiers = [
 ]
 requires-python = ">=3.8"
 dependencies = [
-	"zipp>=0.5",
+	"zipp>=0.5"; python_version < "3.8"',
+	"zipp>=3.19.1"; python_version >= "3.8"',
 	'typing-extensions>=3.6.4; python_version < "3.8"',
 ]
 dynamic = ["version"]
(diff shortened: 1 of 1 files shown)


karistom:
@jaraco please review it, then build and test the change. Thanks.

pyproject.toml
```diff
@@ -18,7 +18,8 @@ classifiers = [
 ]
 requires-python = ">=3.8"
 dependencies = [
-	"zipp>=0.5",
+	"zipp>=0.5"; python_version < "3.8"',
+	"zipp>=3.19.1"; python_version >= "3.8"',
```
jaraco, line 22: Since this library requires Python 3.8+, we can simply declare the dependency once:

```suggestion
	"zipp>=3.19.1",
```

That also fixes the syntax error due to the extra quote character.

jaraco (commented):
I've made one suggestion, but I may not want to accept this change. See the issue for discussion.

karistom:
I updated the zipp clause.

jaraco:
As discussed in #495, this isn't the right approach to take. `importlib_metadata` is compatible with the fixed version and will pick it up by default. It's up to downstream integrators to force the update if needed.

PR #498, opened 2024-08-01
Disallow passing of 'dist' to EntryPoints.select.

jaraco:
Closes python/cpython#107220.


jaraco:
@megies I'd be interested in your opinion on this change (applied first to `importlib_metadata` to later be incorporated into CPython).

megies:
Sorry about the delay. Seems good to me and good to include the proposed use case in the error message  :+1:

PR #501, opened 2024-08-26
Pass mypy and link issues

Avasam:
Ref: https://github.com/jaraco/skeleton/issues/143

Upstream issues:
- https://github.com/jaraco/jaraco.path/issues/4
- https://github.com/python/mypy/issues/11727

### DIFF
--- a/tests/_path.py
+++ b/tests/_path.py
@@ -28,12 +28,12 @@ class TreeMaker(Protocol):
 
 
 def _ensure_tree_maker(obj: Union[str, TreeMaker]) -> TreeMaker:
-    return obj if isinstance(obj, TreeMaker) else pathlib.Path(obj)  # type: ignore[return-value]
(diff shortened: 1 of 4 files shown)


tests/_path.py
```diff
@@ -67,7 +67,8 @@ def build(
 @functools.singledispatch
 def create(content: Union[str, bytes, FilesSpec], path):
     path.mkdir(exist_ok=True)
-    build(content, prefix=path)  # type: ignore[arg-type]
+    # Mypy only looks at the signature of the main singledispatch method. So it must contain the complete Union
+    build(content, prefix=path)  # type: ignore[arg-type] # python/mypy#11727
```
jaraco, line 71: This file is copied from upstream (jaraco.path). These changes will get overridden the next time this code is refreshed. Since they're only comments, let's defer to the comment above to track all issues with this code upstream. (I'll take care of it)

PR #502, opened 2024-08-27
Defer import of zipp

danielhollas:
Impoting zipp takes between 2-4 ms on my machine, which is at least 10% of overall import time of the `importlib_metadata` package (depending on Python version). Since this package is only used in a single code path it makes sense to defer it.

image

To estimate the cost of repeatedly calling `import zipp` I ran

```console
❯ python -m timeit -s "import zipp"  "import zipp"
5000000 loops, best of 5: 59.9 nsec per loop
```
(this is with cpython main branch)

Since this is in a codepath that reads a zipfile from disk, this overhead is negligible.

### DIFF
--- a/importlib_metadata/__init__.py
+++ b/importlib_metadata/__init__.py
@@ -29,8 +29,6 @@ from importlib.abc import MetaPathFinder
 from itertools import starmap
 from typing import Any, Iterable, List, Mapping, Match, Optional, Set, cast
 
-from zipp.compat.overlay import zipfile
-
 from . import _meta
 from ._collections import FreezableDefaultDict, Pair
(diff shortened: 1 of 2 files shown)


jaraco:
I'm starting to feel like a frog being slowly boiled. I'm seeing repeated optimizations like this being handled piecemeal, and each time I do, I need to go through the same process: point out that there's a systemic problem that has [no solution](https://peps.python.org/pep-0690/), ensure there's a comment protecting the change and linking to the motivation, then merge and roll out the change. \</rant>

Looks good. I'll add the comment linking back to the motivation. Thanks for the contrib.

PR #503, opened 2024-08-27
Defer import of json and zipp

danielhollas:
Staged on top of #502 

`platform` and `json` modules take together around 1.5ms to import, around 10% of total import time. Since they are both used in a single place it probably makes sense to inline their import.

After this PR, this is an output of `python -Ximporttime -c "import importlib_metadata"` visualized with tuna using cpython main branch.

image

### DIFF
--- a/importlib_metadata/_compat.py
+++ b/importlib_metadata/_compat.py
@@ -1,3 +1,4 @@
+import platform
 import sys
 
 __all__ = ['install', 'NullFinder']
@@ -51,7 +52,5 @@ def pypy_partial(val):
 
     Workaround for #327.
     """
-    import platform
-
     is_pypy = platform.python_implementation() == 'PyPy'
     return val + is_pypy
--- a/importlib_metadata/_compat.py
+++ b/importlib_metadata/_compat.py
@@ -1,4 +1,3 @@
-import platform
 import sys
 
 __all__ = ['install', 'NullFinder']
@@ -52,5 +51,7 @@ def pypy_partial(val):
 
     Workaround for #327.
     """
+    import platform
+
     is_pypy = platform.python_implementation() == 'PyPy'
     return val + is_pypy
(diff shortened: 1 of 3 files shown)


importlib_metadata/_compat.py
```diff
@@ -52,5 +51,7 @@ def pypy_partial(val):
 
     Workaround for #327.
     """
+    import platform
```
danielhollas, line 54: This seems to fail the code coverage check (diffcov)

https://github.com/python/importlib_metadata/actions/runs/10581007642/job/29317281062#step:5:102

jaraco: That's okay. It's due to there not being coverage aggregation (https://github.com/jaraco/skeleton/issues/130). It's ignorable and will disappear after merging.

jaraco:
Attempting to disentangle the performance impacts of `json` and `platform`:

```
 importlib_metadata main 🐚 .tox/py/bin/python -Ximporttime -c "import importlib_metadata" 2>1 | grep -E r'platform|\sjson$|importlib_metadata$'
import time:        89 |        800 |   json
import time:       221 |        251 |     platform
import time:       636 |      13692 | importlib_metadata
 importlib_metadata main 🐚 800/13692
0.05842827928717499
 importlib_metadata main 🐚 251/13692
0.018331872626351154
```

I see `json` accounts for ~6% of the startup and `platform` under 2% (on my machine under Python 3.12).

importlib_metadata/_compat.py
jaraco, line 54: I'm slightly inclined to say we shouldn't bother moving `platform`. This import only happens in `_compat`, which only happens in zipp (and not in zipfile). Especially given that importing `platform` unconditionally only contributes ~2% to the startup time, I believe the maintenance costs outweigh the benefits. I'll back out this change.

jaraco: Done in 18eb2da0ee267394c1735bec5b1d9f2b0fa77dd9.

Issue #352, opened 2021-09-16
No module 'typing_extensions' on Python 3.7


BKSteve:
Previously posted here: [https://github.com/SynoCommunity/spksrc/issues/4861#issue-995639948](https://github.com/SynoCommunity/spksrc/issues/4861#issue-995639948)

Incompatibility between using Pythhon3 (3.7.11-18) and Python38 (3.8.11-4) in compiled package causes installation failure.

File 
`/var/packages/python3/target/lib/python3.7/site-packages/importlib_metadata/_compat.py`
and
`/var/packages/python38/target/lib/python3.8/site-packages/importlib_metadata/_compat.py`
both have 
```
try:
    from typing import Protocol
except ImportError:  # pragma: no cover
    """
    pytest-mypy complains here because:  error: Incompatible import of "Protocol" (imported name has type "typing_extensions._SpecialForm", local name has type "typing._SpecialForm")
    """
    from typing_extensions import Protocol
```
However Py3 typing.py
`/var/packages/python3/target/lib/python3.7/typing.py` has 
`class _Protocol(Generic, metaclass=_ProtocolMeta):` at line 1168 has underscore prefix and thus cannot be found and thus installation fails.
whereas Py38 typing.py
`/var/packages/python38/target/lib/python3.8/typing.py` has in line with _compat.py `class Protocol(Generic, metaclass=_ProtocolMeta):` is without underscore prefix in front at line 1026 is found and works for installation.

Here's my log of using installed Python 3.7.11-18
```
2021/09/14 12:52:59	Begin service_postinst
2021/09/14 12:53:00	Traceback (most recent call last):
2021/09/14 12:53:00	  File "/var/packages/python3/target/lib/python3.7/site-packages/importlib_metadata/_compat.py", line 8, in <module>
2021/09/14 12:53:00	    from typing import Protocol
2021/09/14 12:53:00	ImportError: cannot import name 'Protocol' from 'typing' (/var/packages/python3/target/lib/python3.7/typing.py)
2021/09/14 12:53:00	During handling of the above exception, another exception occurred:
2021/09/14 12:53:00	Traceback (most recent call last):
2021/09/14 12:53:00	  File "/var/packages/python3/target/bin/virtualenv", line 5, in <module>
2021/09/14 12:53:00	    from virtualenv.__main__ import run_with_catch
2021/09/14 12:53:00	  File "/var/packages/python3/target/lib/python3.7/site-packages/virtualenv/__init__.py", line 3, in <module>
[...]

PR #512, opened 2025-02-03
Add Use Protocol from typing_extensions for Python < 3.8

monou-jp:
No module 'typing_extensions' on Python 3.7 #352

This PR fixes an issue that causes an ImportError when importing importlib_metadata/_meta.py in Python 3.7. Since Protocol does not exist in typing in Python 3.7, this change ensures that it is imported from typing-extensions for Python versions earlier than 3.8.

### DIFF
--- a/importlib_metadata/_meta.py
+++ b/importlib_metadata/_meta.py
@@ -1,17 +1,21 @@
 from __future__ import annotations
 
 import os
+import sys
 from typing import (
     Any,
     Dict,
     Iterator,
     List,
     Optional,
-    Protocol,
     TypeVar,
     Union,
     overload,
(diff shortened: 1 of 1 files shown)


jaraco:
Current releases of `importlib_metadata` only support Python 3.9 and later.  [importlib_metadata 6.7](https://importlib-metadata.readthedocs.io/en/latest/history.html#v6-7-0) is the last version that supported Python 3.7 and [that code has a compatability shim](https://github.com/python/importlib_metadata/blob/f604d3e462cd11ee071bfcb78b827fb67cc2e537/importlib_metadata/_meta.py#L1).

PR #518, opened 2025-04-27
Raise consistent ValueError for invalid EntryPoint.value

### LINKED ISSUE #488: AttributeError changed to AssertionError for invalid identifiers
Since https://github.com/python/importlib_metadata/pull/449 trying to use missing extras raises AssertionErrors. I believe that invalid inputs should never raise AssertionErrors (neither should asserts be used to validate user input).

Observe on Python 3.12 (before this change):

```pycon
>>> ep = importlib.metadata.EntryPoint(name='foo', value='invalid-identifier:foo', group='foo')
>>> ep.extras
  File "<stdin>", line 1, in <module>
  File "/usr/lib64/python3.12/importlib/metadata/__init__.py", line 222, in extras
    return re.findall(r'\w+', match.group('extras') or '')
                              ^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'group'
```

Now this exception message is bogus but at least this is an AttributeError (hence code reading this might assume the .extras attribute is missing).

However, on Python 3.13:

```
>>> ep = importlib.metadata.EntryPoint(name='foo', value='invalid-identifier:foo', group='foo')
>>> ep.extras
  File "<python-input-2>", line 1, in <module>
    ep.extras
  File "/usr/lib64/python3.13/importlib/metadata/__init__.py", line 198, in extras
    assert match is not None
           ^^^^^^^^^^^^^^^^^
AssertionError
```

Existing code (e.g. setuptools) assumes AttributeError in this case:

See https://github.com/pypa/setuptools/blob/544b332bd78d8d274597923e89b9bd7839f8a0f4/setuptools/_entry_points.py#L18-L25

I believe both behaviors are wrong, but the new one is "more wrong".

May I suggest that either `.extras` or `.__init__` raises a specific exception when this happens?

jaraco:
Closes #488

- **Refactored parsing and handling of EntryPoint.value.**
- **Raise a ValueError if no match.**
- **Also raise ValueError on construction if the value is invalid.**
- **Prefer a cached property, as the property is likely to be retrieved at least 3 times (on construction and for module:attr access).**

### DIFF
--- a/importlib_metadata/__init__.py
+++ b/importlib_metadata/__init__.py
@@ -155,6 +155,22 @@ class EntryPoint:
     'attr'
     >>> ep.extras
     ['extra1', 'extra2']
+
+    If the value package or module are not valid identifiers, a
+    ValueError is raised on access.
+
+    >>> EntryPoint(name=None, group=None, value='invalid-name').module
+    Traceback (most recent call last):
+    ...
+    ValueError: ('Invalid object reference...invalid-name...
+    >>> EntryPoint(name=None, group=None, value='invalid-name').attr
+    Traceback (most recent call last):
+    ...
+    ValueError: ('Invalid object reference...invalid-name...
+    >>> EntryPoint(name=None, group=None, value='invalid-name').extras
+    Traceback (most recent call last):
+    ...
+    ValueError: ('Invalid object reference...invalid-name...
     """
 
     pattern = re.compile(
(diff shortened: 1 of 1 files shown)


jaraco:
This change does introduce some performance degradation in the entry_points performance test:

```
===================================== perf =====================================
exercises.py:cached distribution: 0:00:00.000713 (+-1 day, 23:59:59.999988, -2%)
exercises.py:discovery: 0:00:00.000719 (+0:00:00.000002, 0%)
exercises.py:entry_points(): 0:00:00.010800 (+0:00:00.000700, 7%)
exercises.py:entrypoint_regexp_perf: 0:00:00.000070 (+-1 day, 23:59:59.999999, -1%)
exercises.py:uncached distribution: 0:00:00.001090 (+0:00:00, 0%)
```

+.7 ms (7%) in this particular run. I've seen other numbers like 3% on my local machine. I saw a larger increase without the cached property.

This seems to be an acceptable cost.

PR #536, opened 2026-06-19
Prefer valid distributions when collecting entry points

### LINKED ISSUE #534: entry_points doesn't handle empty .dist-info files well
This is related to https://github.com/python/importlib_metadata/issues/489 but with a different way to reach it.

The background is that our build process will delete _files_ when packages are upgraded but will leave _directories_ behind, because it doesn't track who created the directory and whether it should remain or not.

This means that over time you can end up with a `site-packages` that has a number of directories in, such as `setuptools` (with the code), `setuptools-82.0.1.dist-info` (with the 82.0.1 metadata) and also `setuptools-82.0.0.dist-info` which is empty.

If the on-disk order is returning the empty 82.0.0 before 82.0.1 then `importlib_metadata.entry_points()` will return a subset of what is expected.

For example, with a simple test case:

```
for x in metadata.entry_points():
    print(x)
```

A fresh venv that has setuptools/build/pip installed finds 48 entry points:

```
EntryPoint(name='alias', value='setuptools.command.alias:alias', group='distutils.commands')
EntryPoint(name='bdist_egg', value='setuptools.command.bdist_egg:bdist_egg', group='distutils.commands')
EntryPoint(name='bdist_rpm', value='setuptools.command.bdist_rpm:bdist_rpm', group='distutils.commands')
EntryPoint(name='bdist_wheel', value='setuptools.command.bdist_wheel:bdist_wheel', group='distutils.commands')
EntryPoint(name='build', value='setuptools.command.build:build', group='distutils.commands')
...
EntryPoint(name='pip', value='pip._internal.cli.main:main', group='console_scripts')
EntryPoint(name='pip3', value='pip._internal.cli.main:main', group='console_scripts')
EntryPoint(name='pyproject-build', value='build.__main__:entrypoint', group='console_scripts')
EntryPoint(name='build', value='build.__main__:entrypoint', group='pipx.run')
```

But by creating `setuptools-n.dist-info` directories with different values of `n` until `ls -U` shows that it appears _before_ the actual metadata has different behaviour and only finds 4 entry points:

```
EntryPoint(name='pip', value='pip._internal.cli.main:main', group='console_scripts')
EntryPoint(name='pip3', value='pip._internal.cli.main:main', group='console_scripts')
EntryPoint(name='pyproject-build', value='build.__main__:entrypoint', group='console_scripts')
EntryPoint(name='build', value='build.__main__:entrypoint', group='pipx.run')
```

Note that none of the setuptools EPs were listed. 

My theory: the `entry_points()` method is listing all distributions and then doing a `unique()` on it, which will remove duplicate distributions based on the name with the simple logic that it takes the first one seen. With the broken setup I've described above this means it just returns the first, broken, dist and ignores the one with actual content.

I've verified locally that adding another `_prefer_valid()` call to `Distribution.discover()` resolves this by sorting valid dists first:
```
        context = context or DistributionFinder.Context(**kwargs)
        return cls._prefer_valid(itertools.chain.from_iterable(
            resolver(context) for resolver in cls._discover_resolvers()
        ))
```

This feels like a bit of a heavy hammer though and possibly better logic in `entry_points()` would be preferable?

gaoflow:
Fixes #534

## Summary
- add a duplicate-aware distribution selector for `entry_points()` that replaces an invalid/empty dist-info entry with a later valid distribution of the same normalized name
- keep normal discovery order for unique distributions, avoiding the broader cost of preferring valid metadata for every distribution
- add a regression test where an empty older `foo.dist-info` appears before a valid one with entry points

## Tests
- Red before the fix: `PYTHONPATH=. uv run --python 3.11 --with '.[test]' pytest tests/test_main.py::InvalidMetadataTests::test_entry_points_prefer_valid_distribution -q` failed with `KeyError: 'main'`\n- `PYTHONPATH=. uv run --python 3.11 --with '.[test]' pytest tests/test_main.py::InvalidMetadataTests::test_entry_points_prefer_valid_distribution -q`\n- `PYTHONPATH=. uv run --python 3.11 --with '.[test]' pytest tests/test_main.py::InvalidMetadataTests tests/test_api.py::APITests::test_entry_points tests/test_api.py::APITests::test_entry_points_unique_packages_normalized -q`\n- `PYTHONPATH=. uv run --python 3.11 --with '.[test]' pytest -q`\n- `PYTHONPATH=. uv run --python 3.11 --with '.[test,check]' pytest --ruff -q`\n- `PYTHONPATH=. uv run --python 3.11 --with '.[test,type]' pytest --mypy -q`\n- `git diff --check`\n\nAI assistance was used under my direction.

### DIFF
--- a/importlib_metadata/__init__.py
+++ b/importlib_metadata/__init__.py
@@ -1121,6 +1121,21 @@ Wrapper for ``distributions`` to return unique distributions by name.
 """
 
 
+def _unique_valid(dists: Iterable[Distribution]) -> Iterable[Distribution]:
+    """
+    Return unique distributions, preferring valid metadata for duplicates.
+    """
+    has_metadata = ExceptionTrap(MetadataNotFound).passes(operator.attrgetter('metadata'))
+    unique: dict[str, Distribution] = {}
+    for dist in dists:
+        name = dist._normalized_name
+        if name not in unique:
+            unique[name] = dist
+        elif not has_metadata(unique[name]) and has_metadata(dist):
+            unique[name] = dist
+    return unique.values()
+
+
 def entry_points(**params) -> EntryPoints:
     """Return EntryPoint objects for all installed packages.
 
@@ -1131,7 +1146,7 @@ def entry_points(**params) -> EntryPoints:
     :return: EntryPoints for all installed packages.
     """
     eps = itertools.chain.from_iterable(
-        dist.entry_points for dist in _unique(distributions())
+        dist.entry_points for dist in _unique_valid(distributions())
     )
     return EntryPoints(eps).select(**params)
 
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -2,6 +2,7 @@ import importlib
 import pickle
 import re
 import unittest
+from unittest import mock
 
 import pyfakefs.fake_filesystem_unittest as ffs
 from test.support import os_helper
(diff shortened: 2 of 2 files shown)


rossburton:
I've not tested this but I'm not convinced it will actually work in all cases.  Our reason for this is that thanks to package upgrades and removals we'll end up with eg `wheel-0.47.0.dist-info/` existing in `site-packages`, but it will be completely empty.

This is discovered as a distribution, and because `metadata()` falls back to "parse an empty string" it still has valid metadata, it's just that the metadata is empty.  I'm currently hacking a test into pypa/build to check that `dist.metadata.name` is actually set.

rossburton:
I take that back, latest importlib_metadata has different behaviour and correctly raises an exception.

rossburton:
I must admit I don't understand the reasoning behind returning invalid distributions at all: if some empty directories that look like a dist but have no content are found, why return them at all?

binggao1230:
I agree that globally dropping invalid distributions is the cleaner end state if the project wants that behavior. I kept this PR narrower for two reasons:

1. `Distribution.discover()`/`distributions()` is the lower-level discovery surface, and changing it to filter path entries would be a broader API/behavior change than `entry_points()` duplicate resolution.
2. The concrete failure here is narrower: when two distributions normalize to the same name, an invalid/empty older dist-info can win the `unique()` selection and hide the later valid distribution that actually has the entry point. Preferring the valid candidate only inside that duplicate-name choice fixes that failure without changing what raw discovery returns.

So this patch is conservative rather than philosophical: it avoids letting an invalid duplicate shadow a valid package, but does not decide whether invalid dist-info directories should disappear from discovery everywhere. If you prefer the broader policy, I can rework it in that direction.

PR #538, opened 2026-07-13
Add Distribution.editable and top-level editable()

### LINKED ISSUE #510: Add default values to `Distribution.editable` property
Recently, when trying to check if a distribution is editable, I noticed doing so using the current API is very un-ergonomic. This information is available via the `direct_url.json` file of [PEP 610](https://peps.python.org/pep-0610/), which is exposed in `Distribution.origin`.

> When url refers to a local directory, the dir_info key MUST be present as a dictionary with the following key:
>
>    `editable` (type: `boolean`): `true` if the distribution was installed in editable mode, `false` otherwise. If absent, default to `false`.

So, to check if the distribution is editable, we can check `Distribution.origin.dir_info.editable`, however, `Distribution.origin` might be `None` (no `direct_url.json`), and `Distribution.origin.dir_info` might not be set (`url` in `direct_url.json` in not a local directory), which.

```python
is_editable = distribution.origin and distribution.origin.dir_info and distribution.origin.dir_info.editable
```

```python
try:
    is_editable = distribution.origin.dir_info.editable
except AttributeError:
    is_editable = False
```

It would be helpful to have this information available via a more ergonomic API.

apoorva-01:
Closes #510

Adds `Distribution.editable` plus a top-level `editable()`, basically what you sketched in the issue. the property does the null-safe origin.dir_info.editable walk (defaults to False per PEP 610), and editable() just returns distribution(name).editable, same as version()/metadata().

tests cover editable via dir_info, not-editable (origin's there but it's archive_info, no dir_info), and no origin at all. heads up: the existing DistInfoPkgEditable fixture is actually an archive install despite the name, so i added a real dir_info one for the positive case. newsfragment in, full suite + mypy + ruff green.

happy to rename or drop either half if you'd rather go one way.

### DIFF
--- a/importlib_metadata/__init__.py
+++ b/importlib_metadata/__init__.py
@@ -51,6 +51,7 @@ __all__ = [
     'SimplePath',
     'distribution',
     'distributions',
+    'editable',
     'entry_points',
     'files',
     'metadata',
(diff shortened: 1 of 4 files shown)


apoorva-01:
went with the "both" you mentioned, editable(name) just delegates to Distribution.editable (same as version/metadata). property reads origin.dir_info.editable, defaults False when there's no origin/dir_info.

two calls for you: name (editable vs is_editable for a bool?), and both vs just one, happy to drop either.

PR #543, opened 2026-07-16
Make SimplePath path parameters positional-only

### LINKED ISSUE #542: SimplePath rejects pathlib.Path: joinpath and __truediv__ are not positional-only
`SimplePath` describes itself as "A minimal subset of pathlib.Path required by Distribution", but `pathlib.Path` does not satisfy the protocol under ty or pyrefly. Passing a `Path` to `PathDistribution` fails to type check.

## Cause

`joinpath` and `__truediv__` name their parameter `other` without marking it positional-only. Typeshed spells the matching `pathlib.Path` members differently:

```python
# importlib_metadata/_meta.py
def joinpath(self, other: str | os.PathLike[str]) -> SimplePath: ...
def __truediv__(self, other: str | os.PathLike[str]) -> SimplePath: ...

# typeshed stdlib/pathlib/__init__.pyi
def joinpath(self, *other: StrPath) -> Self: ...
def __truediv__(self, key: StrPath) -> Self: ...
```

When a protocol member declares a named parameter, the typing spec requires an implementation to accept that argument by keyword under the same name. `Path.__truediv__` names it `key`, and `Path.joinpath` takes varargs, so neither accepts `other=...`. Both members fail the protocol.

## Reproducer

importlib_metadata 9.0.0, Python 3.14:

```python
import pathlib

from importlib_metadata import PathDistribution

PathDistribution(pathlib.Path('x'))
```

ty 0.0.59 pinpoints it:

```
error[invalid-argument-type]: Argument to `PathDistribution.__init__` is incorrect
 --> repro.py:5:18
  |
5 | PathDistribution(pathlib.Path('x'))
  |                  ^^^^^^^^^^^^^^^^^ Expected `SimplePath`, found `Path`
  |
info: type `Path` is not assignable to protocol `SimplePath`
info: └── protocol member `__truediv__` is incompatible
info:     └── the parameter named `key` does not match `other` (and can be used as a keyword parameter)
```

pyrefly 1.1.1 agrees:

```
ERROR repro.py:5:18-35: Argument `Path` is not assignable to parameter `path` with type
      `SimplePath` in function `importlib_metadata.PathDistribution.__init__` [bad-argument-type]
```

mypy 2.1.0 under `--strict` reports no issues.

I checked each protocol member against `Path` in isolation. `joinpath` and `__truediv__` fail; `parent`, `read_text`, `read_bytes` and `exists` pass.

## Fix

Mark both parameters positional-only, matching how they are called:

```python
def joinpath(self, other: str | os.PathLike[str], /) -> SimplePath: ...
def __truediv__(self, other: str | os.PathLike[str], /) -> SimplePath: ...
```

With that change ty and pyrefly both accept `Path`. mypy keeps passing.

pypa/build hit this while moving off mypy and carries a suppression for it (pypa/build#1135). I will open a PR.

gaborbernat:
Fixes #542.

`SimplePath` names the `joinpath` and `__truediv__` parameter `other` without marking it positional-only, so the typing spec requires an implementation to accept `other=...`. Typeshed names the matching `pathlib.Path` parameters `key` and `*other`, so `Path` fails the protocol under checkers that compare parameter names:

```python
import pathlib

from importlib_metadata import PathDistribution

PathDistribution(pathlib.Path('x'))
```

ty 0.0.59 explains it:

```
info: type `Path` is not assignable to protocol `SimplePath`
info: └── protocol member `__truediv__` is incompatible
info:     └── the parameter named `key` does not match `other` (and can be used as a keyword parameter)
```

pyrefly 1.1.1 reports the same. mypy 2.1.0 accepts it, which is why the docstring's claim that `SimplePath` is "A minimal subset of pathlib.Path" has held up until now.

Adding `/` to both members lets `Path` satisfy the protocol under ty and pyrefly, and leaves mypy passing. I verified all three against importlib_metadata 9.0.0 with the parameters patched, and checked each protocol member against `Path` in isolation: `joinpath` and `__truediv__` were the only failures.

`ruff format`, `ruff check` and `mypy` pass. `tox -e py` fails to collect on this branch, but it fails the same way on a pristine `main` checkout, so that is unrelated.

pypa/build ran into this while moving off mypy (pypa/build#1135).

### DIFF
--- a/importlib_metadata/_meta.py
+++ b/importlib_metadata/_meta.py
@@ -54,11 +54,11 @@ class SimplePath(Protocol):
     """
 
     def joinpath(
-        self, other: str | os.PathLike[str]
+        self, other: str | os.PathLike[str], /
     ) -> SimplePath: ...  # pragma: no cover
 
     def __truediv__(
-        self, other: str | os.PathLike[str]
+        self, other: str | os.PathLike[str], /
     ) -> SimplePath: ...  # pragma: no cover
 
     @property
--- a/newsfragments/542.bugfix.rst
+++ b/newsfragments/542.bugfix.rst
@@ -1,0 +1,1 @@
+Marked the ``other`` parameter of ``SimplePath.joinpath`` and ``SimplePath.__truediv__`` positional-only, so ``pathlib.Path`` satisfies the protocol under type checkers that compare parameter names.


henryiii:
FWIW, I wanted to do this several years ago, when I first found this bug (`Path` and `Traversable` differ only in the name of the argument for `__truediv__`, making them incompatible!). You can see workarounds all over repo-review, which uses `Traversable` heavily. I didn't make it at the time only because Python 3.7 was still supported here. I _think_ mypy added a workaround for this, but it's a hack if so, and other type checkers have not hacked around it. This would fix it at the core: it doesn't matter what a class names the parameter to `__truediv__`, this is passed by position by Python.

gaborbernat:
@jaraco would you have a minute to take a look at this pull request? Thank you.

jaraco:
Looks good to me. My only hesitation - should this be considered a bugfix, feature/general change, or breaking change? On one hand, it is technically backward incompatible and does change behavior. On the other hand, the original intention was that this parameter should always have been positional (and all uses of it in tests are positional). Hyrum's law says someone somewhere is passing the value by keyword. Still, I think those cases are unlikely given this is mainly a protocol meant to constrain the interface of pathlib.Path (not expand it). Bugfix is the right choice.

gaborbernat:
I'd imagine it would break anyway at runtime, no? If they use that?

